PR #24, opened 2015-06-24
Attn: Luke; Subject: i18n translations with BOOK-788 

gwenau:


gwenau:
This has pull request has been replaced with pull request #120.

PR #48, opened 2015-09-29
Admin Dashboard

simonrice:
This pull request contains a few changes:

* Some fixes that allow the SDK to be used in our admin dashboards - in the end, we need the company path in the scope, as well as adding the auth token to the initial request.
* It also corrects the spelling of DateTimeUtilitiesService (sorry, that one was annoying me!)


edclements:
Why do you need the company api path in scope? Surely the company already is.

simonrice:
Closing as this won't be used (& it's sticking out like a sore thumb to me!)

PR #63, opened 2015-11-05
passing raw mobile if provided

undersky0:


lukeferguson:
Phone number stuff is getting a bit messy - really formatting should be the responsibility of a $formatter method that's associated with phone number inputs, thus there should no need to assign the raw_mobile as the mobile. I've raised PLAT-31to address this.

lukeferguson:
No need this anymore - `raw_mobile` has been removed from the SDK as of v0.3.0.

PR #68, opened 2015-11-10
[CES-18] Standard Angular Admin Events Journey

gwenau:
Main changes:
- momentjs dependency added
- admin events api updated to match admin_api.rb
- client_admin page styled
- some fields hidden as it does not apply to an admin user


lukeferguson:
Hey @gwenau 
`event = halClient.$get($scope.bb.api_url + '/api/v1/admin/' + company_id + '/event_chains/' + $scope.bb.item_defaults.event_chain + '/events/' + $scope.bb.item_defaults.event )` looks a bit dodgy - is that valid api method?

Also I've noticed there's a reference to blink in `bb-sass-3-3-1/_variables.scss`.

lukeferguson:
I stand corrected, it is a api method! Regarding teh font assets, these should be loaded from the rails assets - they appear to be there already :smile:

gwenau:
precss will have to be applied to the bb gulpfile as a different font path is required for the bespoke journeys. In the meantime, the font path is hardcoded to a bespoke path.

PR #72, opened 2015-11-13
Scope.bb.basket.settings.requires deal

WesleyPumpkinhead:
Added $attrs dependancy to controller BasketList so that requires_deal can be assigned to bb.basket.settings using bb-basket-list='{requires_deal:true}'.

Updated setReady() so it now fails if settings.requires_deal is true and hasDeal() returns false (do not proceed to checkout).

Added a new error object to errors.js.coffee - "GIFT_CERTIFICATE_REQUIRED"

Resolved merge conflict by removing ErrorService dependancy and adding $attrs dependancy from/to BasketList controller and using the new AlertService.raise(<error_key>) method for "GIFT_CERTIFICATE_NOT_FOUND"

### DIFF
--- a/src/core/javascripts/controllers/basket.js.coffee
+++ b/src/core/javascripts/controllers/basket.js.coffee
@@ -60,11 +60,15 @@ angular.module('BB.Directives').directive 'bbBasketList', () ->
   controller : 'BasketList'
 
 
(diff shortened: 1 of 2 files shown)


lukeferguson:
Looks good @WesleyPumpkinhead! A couple of sneaky console logs are in basket.js.coffee though, can you remove those and I'll then merge it :smile:

WesleyPumpkinhead:
Console Logs removed @lukeferguson

PR #75, opened 2015-11-17
ngdocs CRUD for models second part

robua:


edclements:
Can we keep the indentation consistent please. It should be two spaces

robua:
I've checked the indentation, it's two spaces. Please tell me which file(s) you saw  wrong identation. Please review PR and advise if CRUD is correctly implemented for models. Thanks.

src/admin/javascripts/models/booking.js.coffee
```diff
@@ -76,3 +164,37 @@ angular.module('BB.Models').factory "Admin.BookingModel", ($q, BBModel, BaseMode
...
+    * @param {integer=} per_page Number of items per page of paginated response.
+    * @param {array} include_cancelled An array with booking list what include cancelled booking.
+    * @methodOf BB.Models:AdminBooking
+    * @description
+    * Gets a filtered collection of bookings.
+    *
+    * @returns {Promise} Returns a promise that resolves to the filtered collection of bookings.
+    ###
+    @query: (company, slot_id, start_date, end_date, service_id, resource_id, person_id, page, per_page, include_cancelled) ->
+      AdminBookingService.query
```
Glenn: I asked this before - but maybe have missed that answer - does this actually work ? I don't see the inclusion of AdminBookingService via DI 
I would be good to see some tests for these - even if they use fake end points

src/admin/javascripts/models/slot.js.coffee
```diff
@@ -20,3 +33,76 @@ angular.module('BB.Models').factory "Admin.SlotModel", ($q, BBModel, BaseModel,
...
+    * @param {integer=} page Specifies particular page of paginated response.
+    * @param {integer=} per_page Number of items per page of paginated response.
+    * @methodOf BB.Models:AdminSlot
+    * @description
+    * Gets a filtered collection of slots.
+    *
+    * @returns {Promise} Returns a promise that resolves to the filtered collection of slots.
+    ###
+    @query: (company, start_date, end_date, date, resource_id, service_id, person_id, page, per_page) ->
+      AdminSlotService.query
```
Glenn: Same here, is the AdminSlotService loaded ?

PR #85, opened 2015-11-27
[SKGY-22] Moved the move booking alert logic into its own function so…

DeanPalmer:
… that it can be called from both the appointment and event move booking logic.


lukeferguson:
I'd rename the method to showMoveMessage or raiseMovedAlert @DeanPalmer, otherwise looks good :smile:

PR #92, opened 2015-12-16
[PLAT-21] Added new classes to each calendar day row. Using these mad…

Morphasis:
Purpose of this pull request is to add new calendar resizing functionality.

It is now set up in this template to now go from 7 columns on large devices to 5 on medium to small devices then goes to 1 extra small devices. This improves the readability and usability of the calendar.

Doing this renders the class added on day.hide unless so i has been removed.

Additionally there is some custom padding added when it resizes so the calendar lines up better on smaller devices.


src/public-booking/stylesheets/_calendar.scss
```diff
@@ -159,6 +159,32 @@
     }
   }
 
+// calendar hidden column functionality
+.cal-col-2, .cal-col-3, .cal-col-4, .cal-col-5{
+  @media (max-width: 768px) {
+    display: none;
```
lukeferguson: screen-sm-max var should be used here

lukeferguson:
See comment on _calendar.scss.

Also, I'd update the bbTimeRanges directive to remove updateHideStatus (and calls to it) as this now redundant.

PR #95, opened 2015-12-17
[DEBS-183] Added default redirectTo method for the widget.

Morphasis:
This pull request is to add a new total app based redirectTo method. Then removed it from the directive that had this method defined as it was not unneeded.

### DIFF
--- a/src/core/javascripts/controllers/base.js.coffee
+++ b/src/core/javascripts/controllers/base.js.coffee
@@ -16,7 +16,7 @@
 * restrict: 'A'
 * scope:
 *   client: '=?'
-*   apiUrl: $
+*   apiUrl: '@?'
 *   useParent:'='
 * transclude: true
 * </pre>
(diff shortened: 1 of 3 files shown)


src/core/javascripts/controllers/base.js.coffee
```diff
@@ -16,7 +16,7 @@
 * restrict: 'A'
 * scope:
 *   client: '=?'
-*   apiUrl: '@?'
+*   apiUrl: $
```
lukeferguson: Sorry @Morphasis - I accidently changed this - can you can change it back please

src/core/javascripts/controllers/base.js.coffee
```diff
@@ -1367,3 +1367,8 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location,
   $scope.scrollTo = (id) ->
     $location.hash(id)
     $anchorScroll()
+
+
+  $scope.redirectTo = (url) ->
+    debugger
```
lukeferguson: Can you remove this @Morphasis

PR #98, opened 2015-12-17
JES-2

WesleyPumpkinhead:
Improved functionality of month picker.
1. Dates that are not disabled will no longer do anything if clicked on. Also, they will not have circles.
2. Dates that are NOT disabled but are NOT available will behave as per above.


src/core/javascripts/controllers/event_list.js.coffee
```diff
@@ -418,15 +441,15 @@ angular.module('BB.Controllers').controller 'EventList', ($scope, $rootScope, Ev
...
       (($scope.filters.event_group and item.service_id == $scope.filters.event_group.id) or !$scope.filters.event_group?) and 
       (($scope.filters.price? and (item.price_range.from <= $scope.filters.price)) or !$scope.filters.price?) and
       (($scope.filters.hide_sold_out_events and item.getSpacesLeft() != 0) or !$scope.filters.hide_sold_out_events) and
       filterEventsWithDynamicFilters(item)
-    return result
-
+    return result  
 
+  # I had to put this into $scope so that MultiEventList controller (which extends this controller) could overwrite $scope.filterEvents
   filterEventsWithDynamicFilters = (item) ->
```
lukeferguson: @WesleyPumpkinhead comment suggests this should be scope, but it hasn't been changed? Either update the method or remove the comment - whichever is correct :)

src/core/javascripts/controllers/event_list.js.coffee
```diff
@@ -248,14 +259,21 @@ angular.module('BB.Controllers').controller 'EventList', ($scope, $rootScope, Ev
...
-
-      $scope.items = _.flatten(_.toArray($scope.events))
+     
+      # Flatten events array
+      $scope.items = _.flatten(events)
+     
+      # Add spaces_left prop - so we don't need to use ng-init="spaces_left = getSpacesLeft()" in the html template      
+      for item in $scope.items
+        item.spaces_left = item.num_spaces - item.getNumBooked()
+
```
lukeferguson: You should just call item.getSpacesLeft() (a method on the event model) rather than doing the calculation yourself.

lukeferguson:
@WesleyPumpkinhead, nice work :+1:   I've made a couple comments, however, I've still merged the pull request due to their triviality and the fact I need you fix to showDay too!

PR #101, opened 2015-12-21
invalid email error handling and regex

undersky0:
Invalid error handling and error message. Also new regex for front end password validation.


lukeferguson:
@undersky0 - is the pattern the standard devise password pattern?  If so, can you please rename the method and password property to getStandardPassword Pattern / standard_password  - a little more descriptive than password81 I think.

undersky0:
yes that looks better

PR #113, opened 2016-01-13
Restrict lodash.js to version 3.10.1 as version 4.0.0+ renames variou…

WesleyPumpkinhead:
…s underscore.js methods and breaks everything


WesleyPumpkinhead:
@lukeferguson

PR #114, opened 2016-01-14
Bb background image

WesleyPumpkinhead:


src/core/javascripts/directives/main.js.coffee
```diff
@@ -209,6 +195,34 @@ angular.module('BB.Directives').directive 'bbModal', ($window, $bbug) ->
...
+* Adds a background-image to an element
+
+* @param
+* {string} url
+*
+* @example
+* <div bb-background-image='images/example.jpg'></div>
+####
+angular.module('BB.Directives').directive('bbBackgroundImage', () ->
+  return{
```
lukeferguson: No need for the directive to wrapped in a return @WesleyPumpkinhead

WesleyPumpkinhead: But there is! It's just that coffeescript does it for you =P

PR #117, opened 2016-01-14
Browser back button fix to remove pages from history

WesleyPumpkinhead:
Added new functionality to routing. Angular's $locationChangeStart broadcast now checks if we are going back in the journey steps and calls the loadPreviousStep method - which has been updated to remove pages from the browser history using JavaScript's native window.history.go method. I have also updated the matchURLToStep method in widget.js to use the allSteps array instead of the dynamically generated steps array.


src/core/javascripts/controllers/base.js.coffee
```diff
@@ -702,13 +702,23 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location,
...
 
-  $scope.$on '$locationChangeStart', (event) =>
-    return if !$scope.bb.routeFormat
-    if !$scope.bb.routing
-      step = $scope.bb.matchURLToStep()
-      $scope.loadStep(step) if step
-    $scope.bb.routing = false
 
+  # $locationChangeStart is broadcast before a URL will change
+  $scope.$on '$locationChangeStart', (angularEvent, newUrl, oldUrl) ->
```
lukeferguson: arguments and variables should always use under_score notation

src/core/javascripts/controllers/base.js.coffee
```diff
@@ -1164,25 +1174,36 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location,
   * @description
   * Loads the previous unskipped step
   *
-  * @param {object} steps_to_go_back The number of steps to go back
+  * @param {integer} steps_to_go_back: The number of steps to go back
```
lukeferguson: I'm happy with steps_to_go_back to be deleted, but we need to ensure we don't break the debenhams widget, the reason this functionality was introducted - here I was trying to go back two steps to avoid going back to the login page. However, what I should have done was scall skipThisStep() in bbMemberLogin in the handleLogin method just before decideNextPage is called.  This will ensure users aren't forced to login in again going back or routed forward automatically.

lukeferguson:
Reviewed @WesleyPumpkinhead - couple to changes to make as discussed.

WesleyPumpkinhead:
@lukeferguson - Tested with debenhams and you are no longer routed to the login page once logged in. One thing I noticed is that the $loadPreviousStep method is called when you select one of the appointment types from the index.html.

In the address bar it looks like:

    index.html
    personal_shopper.html
    (then $scope.loadPreviousStep is called)
    personal_shopper.html#/shopper_service_list

but this happens with or without my changes, and regardless of whether you are signed in or not.

PR #119, opened 2016-01-18
[PLAT-39] Moved member model to core to fix the error on view booking…

Morphasis:
Fix: 
This fix is to stop people requiring member as a dependency when the view booking email pages require member by default so it was moved to core.


lukeferguson:
No need for this change as discussed yesterday @Morphasis . We should instead check if the model exists before attempting to instantiate it.

PR #120, opened 2016-01-18
SMED-28 Update standard widget to support i18n

gwenau:
In the examples folder, a service journey that can be tested is 37002 https://mamasandpapas-staging.bookingbug.com. And an event journey that can be tested is 37000 https://bmgf-dev.bookingbug.com.


gwenau:
Replaced with pull request #121 as the new PR at least passes the local gulp test.

PR #127, opened 2016-01-27
[PLAT-41] Update calendar template to include choice of 3 different c…

WesleyPumpkinhead:
…alendars with descriptions. Fixed bug with calendar button not closing calendar when clicked. Added new default neutral styles for ui-calendar which will be more difficult to accidentally override when applying custom styles, in particular with respect to the bootstrap btn classes


src/public-booking/templates/calendar.html
```diff
@@ -59,8 +136,8 @@ <h2 class="month-heading">{{pretty_month_title('MMMM', 'YYYY')}}</h2>
             </button>
           </span>
           <h3 class="day-heading col-xs-8 col-sm-12">
-            <span class="visible-xs">{{day.date | datetime: "ddd"}} <strong>{{day.date | datetime: "D"}} </strong>{{day.date | datetime: "MMM"}}</span>
-            <span class="hidden-xs">{{day.date | datetime: "ddd"}} <strong>{{day.date | datetime: "D"}} </strong></span>
+            <span class="visible-xs">{{day.date | datetime: "ddd":false}} <strong>{{day.date | datetime: "D":false}} </strong>{{day.date | datetime: "MMM":false}}</span>
```
lukeferguson: There's need for :false for datetime objects with @Morphasis recent changes @WesleyPumpkinhead

WesleyPumpkinhead:
@lukeferguson obsolete false args removed from datetime filter

PR #129, opened 2016-01-28
Bookingbug angluar ui grid

robua:


edclements:
This is looking good but can you make sure all instances of tr-ng-grid are replaced and remove the module being loaded.

PR #133, opened 2016-01-29
Add new _mixins.scss to public-booking

WesleyPumpkinhead:


src/public-booking/stylesheets/main.scss
```diff
@@ -56,4 +57,5 @@ $secondary-border-color     : #D5D5D5;
   @import "time";
   @import "filters";
   @import "admin";
+  @import "popovers";
```
lukeferguson: Hey @WesleyPumpkinhead  - doesn;t look like this file is included?

PR #135, opened 2016-02-02
[MOTH-22][PART1-API] attach answers to item_details.questions

123gpg321:
Obscure bug fix in 2 parts. SDK and API. Might need revisiting at some point.


edclements:
Should asked_questions be skipped? Maybe it's just not being set correctly somewhere else.

123gpg321:
Will check with @lukeferguson

PR #147, opened 2016-02-25
[PLAT-52] The tags are swapped for play 51 and 2 this is the commit f…

Morphasis:
This is a template update for the new client_admin. Ported over from the capone journey using standard variables now from variable.scss.

The new features are:
1.) Typehead (This javascript has already been included in the sdk i made this earlier) support api requests on type so that it does not put to much strain on the journey.
2.) The old click search functionality and get a table is still there but entirely remade and looks better. Simply type something in the Typehead then hit search icon to get the old table information in a new format.

- Kind of breaking changes - 
1.) Font awesome is required for the icons this should be included by default in the sdk soon. If it is not at the time of reading this.

- Usage -
1.) Copy over the template from public booking client_admin
2.) Copy over _client_admin scss file
3.) Template now works but you need to setup the relevant dash settings just like you had to prior. Or pass in the qs yourself.


Morphasis:
DO NOT MERGE
This is currently waiting on a integration of the partial render in it. I just saw this i will add it and remove this comment when i have time to update this.

PR #153, opened 2016-03-03
Fix timezone issue with event date

WesleyPumpkinhead:


WesleyPumpkinhead:
@lukeferguson parseZone method and event.js tidy up as discussed yesterday evening

PR #157, opened 2016-03-07
[SMED-28]: Build i18n into SDK without breaking backwards-compatibility

tomlandia:
# Main changes:
- Replaces all hardcoded template and filter text with angular-translate keys
- Uses `messageformat` interpolation if the string needs pluralization in any language
- Standarises moment formats (using `LL*`, `LT` as much as possible)
- Adds `print_local_time` method to `time_slot model`
- Fixes currency and time period filters
- Lazy-loads non-English moment locales and translation keys in i18n JS files packaged in core bower module.
- Adds category_list.html and monthly_calendar.html public booking templates [PLAT-64]
- Replaces `moment` fork dependency by a `toIsoDate` polyfill
- Gets rid of 200kB `angular-google-maps` dependency which was used only for the address map

To use:
- Put your custom translation keys in `src/i18n/`
- Set `i18n` `bb-widget` param to `true` and `supported_locales` param to the locales you want to support (you can leave this param if you don't use custom keys)
- Add `bb-locale-switcher` directive in your main view so the user can switch languages
- Use `messageformat` interpolation if you need pluralization or genderization (it's not compatible with angular fitlers)
- If necessary, update your project's gulpfile if so your files get merged with core ones

Limitations:
- German keys missing
- Gulpfile needs to be updated in client projects as the files have to be lazyloaded
- `bb-locale-switcher` directive only included in public-booking by default
- All API data such as activity names and questions are obviously English-only
- Phone number and address filters are still UK/US-specific
- No distinction between American and British English (eg date format)
- Datepicker date formats aren't internationalized
- Bower conflict because we _need_ `messageformat` v0.3.1 even if `angular-translate` wants v0.16 because v0.16 requires bloating json locale files.

Tips for SDK developers:
- Use `translate-interpolation="messageformat"` if you need pluralization or genderization
- Please do not break sentences into several keys
- Use `LL/LLL/LLLL/LT` moment formats whenever possible


Morphasis:
@lukeferguson Luke dont forget to uncomment out this when merging this pull request 
https://github.com/BookingBug/bookingbug-angular-bespoke/commit/15540ef658b7a95709dd76e9f9cad0c0c5c2b72b
-JS

PR #159, opened 2016-03-08
[PLAT-38] Improve calendar accessibility

tomlandia:
Branched from SMED-28-2 so to be merged after #157.


tomlandia:
Replaced by #160 as ngAria makes these changes useless

PR #160, opened 2016-03-08
[PLAT-38] [PLAT-57] Accessibility improvements

tomlandia:
Fixes PLAT-38 and works towards PLAT-57

Branched from SMED-28-2 to make labels multilingual so to be merged after #157.


tomlandia:
Replaced with #307 (based on current master)

PR #161, opened 2016-03-09
[PLAT-42] add jumpToNextAvailablePeriod option to bbTimesRange

tomlandia:
Should be merged after #157. I branched from SMED-28-2 to avoid lodash/underscore conflicts.


tomlandia:
Seems to have been implemented elsewhere.

PR #170, opened 2016-03-15
Replaced glyphicons with fontawesome, removed alerts (these should no…

WesleyPumpkinhead:
…t be in this template), replaced ng-style directive for background image with bb-background-image directive and updated styles accordingly


WesleyPumpkinhead:
@lukeferguson 

http://take.ms/rRluo
http://take.ms/17c6v

lukeferguson:
@WesleyPumpkinhead the info icon feels a bit meaningless for duration... maybe make use the glyphicons-stopwatch or find an altenatifve to add to bb-icons?

WesleyPumpkinhead:
@lukeferguson I have replaced the info icon with hourglass-half icon which is available in latest version of fontawesome

PR #172, opened 2016-03-18
[FRAM-145] getting purchases

undersky0:


lukeferguson:
hey @undersky0, how does removing from the connection_started promise make this work?

PR #178, opened 2016-03-23
PLAT-62 adding ability to block events

lefterisk:


src/admin-dashboard/javascripts/directives/resource_calendar.js.coffee
```diff
@@ -101,6 +105,8 @@ angular.module('BBAdminDashboard').directive 'bbResourceCalendar', (
           rid = resource.id if resource
           $scope.getCompanyPromise().then (company) ->
             AdminBookingPopup.open
+              fromDatetime: start
+              toDatetime: end
```
lukeferguson: var names should be underscore notation

PR #180, opened 2016-03-24
PLAT-62 adding a resource picker to the block section

lefterisk:
For the week and month view that dont have a pre selected resource

### DIFF
--- a/src/admin-booking/javascripts/directives/block-time.js.coffee
+++ b/src/admin-booking/javascripts/directives/block-time.js.coffee
@@ -32,11 +32,11 @@ angular.module('BBAdminBooking').directive 'bbBlockTime', () ->
 
     # On select change update the right current_item variable depending 
     # whether the selected item is a person or a resource
-    $scope.$watch 'picked_resource', (newValue, oldValue)->
-      if newValue != oldValue and newValue?
-        parts = newValue.split '_'
+    $scope.changeResource = ()->
+      if $scope.picked_resource?
+        parts = $scope.picked_resource.split '_'
         angular.forEach $scope.resources, (value, key)->
-          if value.identifier == newValue
+          if value.identifier == $scope.picked_resource
             if parts[1] == 'p'
               $scope.bb.current_item.person = value
             else if parts[1] == 'r'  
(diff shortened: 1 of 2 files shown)


Glenn:
Can we call people staff or people, and not "resources"

Also - can we extend this also also support actual resources (and detect if they aren't configured)

lefterisk:
Added the ability to select resource as well as a person. Both lists are merged into the same select which has OptionGroups 'Staff' and 'Resources'

src/admin-booking/templates/quick_pick.html
```diff
@@ -25,15 +25,14 @@
...
     <h1>Block Time</h1>
-
     <div class="form-group">
       <label>For</label>
-      <p ng-show="bb.current_item.person.name">
-        {{bb.current_item.person.name}}
-      </p>
       <p ng-show="bb.current_item.resource.name">
         {{bb.current_item.resource.name}}
       </p>
```
lukeferguson: this can be deleted too

lefterisk: Done

src/admin-booking/javascripts/directives/block-time.js.coffee
```diff
@@ -2,6 +2,47 @@ angular.module('BBAdminBooking').directive 'bbBlockTime', () ->
...
+    # If in "Day" view a person or resource will have been passed in
+    if $scope.bb.current_item.person? and $scope.bb.current_item.person.id?
+      $scope.picked_resource = $scope.bb.current_item.person.id + '_p'
+
+    if $scope.bb.current_item.resource? and $scope.bb.current_item.resource.id?
+      $scope.picked_resource = $scope.bb.current_item.resource.id  + '_r' 
+
+    # On select change update the right current_item variable depending 
+    # whether the selected item is a person or a resource
+    $scope.$watch 'picked_resource', (newValue, oldValue)->
```
lukeferguson: use of ng-change bound to the select control would be appropriate - $watch gets evaluated every digest cycle whereas ng-change is only invoked when the drop down changes

lefterisk: Swapped $watch with ng-change

PR #195, opened 2016-04-05
[LHT-102] don't set Authorization header to undefined

tomlandia:


src/core/javascripts/components/angular-hal.js
```diff
@@ -330,11 +330,12 @@ angular
...
-        'Authorization': options.authorization
-        , 'Content-Type': 'application/json'
-        , 'Accept': 'application/hal+json,application/json'
-      }
+      var headers = {
+        'Content-Type': 'application/json',
+        'Accept': 'application/hal+json,application/json'
+      };
+      if(options.authorization)
+        headers.Authorization = options.authorization;
```
lukeferguson: @tomlandia just put it all in a single line akin to the two statements below

tomlandia: Fixed

PR #196, opened 2016-04-05
Fix member sso login

WesleyPumpkinhead:


WesleyPumpkinhead:
@jackbryant I had to make some changes to the memberSsoLogin directive. The TODO? comment is Luke's. Do we need to set the client to the member? I think there is a method setClient(memberObject) used elsewhere.

Please see me tomorrow to discuss and explain how I cant test once you have merged your API changes.

PR #203, opened 2016-04-07
[PLAT-63] Adding airbrake environment check

Grigore-George-Mihai:


lukeferguson:
hey @Grigore-George-Mihai - you're only modifying the payload there, in this case we actually want to stop errors posting when running localhost: e.g.

```
airbrake.addFilter (notice) ->
  if (AirbrakeConfig.environment is "development"
    // Ignore errors from development
    return null
  return notice
```

then set the config like this:

```
 app.value('AirbrakeConfig', {	
    projectId: '34693'
   projectKey: 'b4174c79b8b7dfb0111e45aa35c95b71','b4174c79b8b7dfb0111e45aa35c95b71'
   environment: if window.location.hostname is "locahost" then "development" else "production"
  })

```

PR #206, opened 2016-04-11
Fixes problem with cancellation modal not always being on top of othe…

WesleyPumpkinhead:
…r content. It currently has a z-index of about 1040, and this is not enough because a client's custom content such as headers may exceed this


lukeferguson:
Hey @WesleyPumpkinhead, this might be better to apply this fix to the  bbModal directive - this directive already takes of care of some sizing issues and should be applied to all modals.

PR #214, opened 2016-04-18
[PLAT-68] end-to-end tests & unit testing

tomlandia:
Adds Protractor end-to-end tests. Uses SauceLabs on TravisCI as using Chrome directly on TravisCI wouldn't work and using PhantomJS would be too restrictive.

For the `/checkout` mockup, I had to add an ugly script to the HTML example as I really can't get `browser.addMockModule` to work and higher-level solutiond like `httpbackend` and `protractor-http-mock` rely on `browser.addMockModule`.


maxmckenzie:
@tomlandia so what steps would i need to do to get this running locally? do i need any global deps?

have you got a link to the get `browser.addMockModule` docs?

tomlandia:
Everything should be installed automatically. http://www.protractortest.org/#/api?view=Protractor.prototype.addMockModule . First you'd need to get the test to work with Angular synchronisation enabled.

tomlandia:
Everything should be installed automatically (run `npm install` and `gulp e2e-tests`.

http://www.protractortest.org/#/api?view=Protractor.prototype.addMockModule

PR #221, opened 2016-04-20
[LTH-94] putting core dependencies in the correct bower.json + preemp…

lefterisk:
…tive work for change in the booking api endpoint


edclements:
It would be good to ensure where the booking service is being called, that the company is being passed in. Those hard coded URLs are a bit of a code smell.

PR #225, opened 2016-04-25
[ORIENTAL-66] this superfluous call retriggers a watcher which eventu…

lefterisk:
…ally reverts to the previous day


edclements:
Perhaps it should just be checking if currentDate is not already defined?

lefterisk:
It always going to be set, https://github.com/BookingBug/bookingbug-angular/blob/oriental-66/src/admin-dashboard/javascripts/directives/resource_calendar.js.coffee#L258

PR #246, opened 2016-05-05
[ORIENTAL-80] Pointed to the correct modal.

DeanPalmer:


DeanPalmer:
Changes the modal that appears when a booking is clicked. Tested and all functionality works as expected.

PR #249, opened 2016-05-09
[SKYG-108] calendar .slick-hidden class fix

victorjeman:


src/public-booking/stylesheets/_calendar.scss
```diff
@@ -284,6 +284,15 @@
 
   }
 }
+
+// Show the arrows but make them grey as in .slick-disabled
+.slick-arrow.slick-hidden {
```
lukeferguson: hey @victorjeman, calendar.scss is for the week calendar component (bbTimeRanges), this css should be in event_list.scss

PR #266, opened 2016-05-18
[BW-88] experimental use_company_timezone param

tomlandia:
`use_company_timezone` (default false) widget param makes the widget display times in local timezone rather than in the business'. The dates without a time remain unchanged.

I deleted the hardcoded moment formats in the filter as no project still uses them.


src/core/javascripts/filters/main.js.coffee
```diff
@@ -282,26 +282,15 @@ angular.module('BB.Filters').filter 'local_phone_number', (SettingsService, Vali
...
-      datestrings =
-        datetime_us : 'MM/DD/YYYY, h:mm a'
-        datetime_uk : 'DD/MM/YYYY, HH:mm'
-        date_us : 'MM/DD/YYYY'
-        date_uk : 'DD/MM/YYYY'
-        time_us : 'h:mm a'
-        time_uk : 'HH:mm'
-      cc = SettingsService.getCountryCode()
-      cc = "uk" if cc != "us"
-      if format and format.match(/(date(time_uk|time_us|_us|_uk)*|(time(_uk|_us)*))/)
```
lukeferguson: @tomlandia we do want to preserve this functionality - it's useful when you quickly have just the date or time quickly formatted.  Obviously, LLL should be the default format is none is provided

src/core/javascripts/filters/main.js.coffee
```diff
@@ -282,26 +282,28 @@ angular.module('BB.Filters').filter 'local_phone_number', (SettingsService, Vali
...
+  local_time_regex = /LT|H/i
+  local_utc_offset = moment().utcOffset()
+  hardcoded_formats =
+    datetime_us : 'MM/DD/YYYY, h:mm a'
+    datetime_uk : 'DD/MM/YYYY, HH:mm'
+    date_us : 'MM/DD/YYYY'
+    date_uk : 'DD/MM/YYYY'
+    time_us : 'h:mm a'
+    time_uk : 'HH:mm'
+  hardcoded_formats_regex = /date|time/
```
lukeferguson: @tomlandia need to test for `datetime` too

src/core/javascripts/filters/main.js.coffee
```diff
@@ -282,26 +282,28 @@ angular.module('BB.Filters').filter 'local_phone_number', (SettingsService, Vali
...
+
+  (date, format="LLL") ->
+    if format.match(hardcoded_formats_regex)
+      cc = SettingsService.getCountryCode() || 'uk'
+      format = hardcoded_formats["format_#{cc}"]
```
lukeferguson: `format` needs to be interpolated too

src/core/javascripts/filters/main.js.coffee
```diff
@@ -282,26 +282,28 @@ angular.module('BB.Filters').filter 'local_phone_number', (SettingsService, Vali
...
-        date_uk : 'DD/MM/YYYY'
-        time_us : 'h:mm a'
-        time_uk : 'HH:mm'
-      cc = SettingsService.getCountryCode()
-      cc = "uk" if cc != "us"
-      if format and format.match(/(date(time_uk|time_us|_us|_uk)*|(time(_uk|_us)*))/)
-        return date.format(datestrings[format + "_" + cc])
-      else if format
-        return date.format(format)
+      if SettingsService.getUseLocalTimezone() and format.match(local_time_regex)
```
lukeferguson: Is it necessary to check both SettingsService and the format here?

tomlandia: This is a hack to avoid changing dates if utcOffset < 0 (all dates are set up at midnight).

PR #268, opened 2016-05-19
@bulk_purchase added to basket_item model

gwenau:
@bulk_purchase added to isTimeItem() in basket_item model for bbBasketList directive to load properly.


gwenau:
Found a more suitable solution.

gwenau:
Looks like it's required to be able to retrieve prepaid bookings.

lukeferguson:
@gwenau time items should only be events and appointments - message me on slack to discuss what you're trying to achieve here

PR #274, opened 2016-05-19
[ORIENTAL-111] Fix overlay of schedules on week view

lefterisk:


edclements:
In the future this could be re-factored into a service

PR #289, opened 2016-05-26
[AVE-18] Show booking moved date with US date format condition

gwenau:
At the moment, the moved booking date is only displaying in UK format. An extra condition has been created to display it in US format.

### DIFF
--- a/src/core/javascripts/controllers/item_details.js.coffee
+++ b/src/core/javascripts/controllers/item_details.js.coffee
@@ -270,10 +270,7 @@ angular.module('BB.Controllers').controller 'ItemDetails', ($scope, $attrs, $roo
       $translate('MOVE_BOOKINGS_MSG', { datetime:datetime.format('LLLL') }).then (translated_text) ->
         AlertService.add("info", { msg: translated_text })
     else
-      cc = SettingsService.getCountryCode()
-      cc = "uk" if cc != "us"
-      date_format = if (cc == 'uk') then 'LLLL' else 'dddd, MMMM Do[,] h:mm A'
-      AlertService.add("info", { msg: "Your booking has been moved to #{datetime.format(date_format)}" })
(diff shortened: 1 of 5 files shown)


tomlandia:
We should always use `L*` formats instead of having conditionals everywhere. Please edit `src/core/javascripts/main.js.coffee` instead.

lukeferguson:
@gwenau see comment from @tomlandia above

gwenau:
Hi @tomlandia,

I attempted to modify the src/core/javascripts/main.js.coffee file but was not getting any results. Calling the country code at that stage only returned an empty string so I added the format condition to the basket_item model.

Pros:
- Potentially, we can set the right US/UK date format here and that could mean that we rely less on the datetime filter

Cons:
- This code will run each time there is a datetime object so it might slow done the processing speed a bit (compared to the previous commit for this pull request)

tomlandia:
We can get rid of the con by moving this code to SettingsService.setCountryCode.

gwenau:
Con has been removed by moving code to SettingsService.setCountryCode.

tomlandia:
Make sure to revert the first commit (f36ece3).
Would be good to do that for all `L*` formats. Luke suggests even embedding [en-gb.js](https://github.com/moment/moment/blob/develop/locale/en-gb.js) and calling `moment.locale('en')`/`moment.locale('en-gb')` (I think American is the default)

gwenau:
I've embedded the moment locale 'en-gb' file in core/javascripts/services/i18n. This seems to be the most logical place to put it. Unless the company code is 'us', the date will be returned in the British format.

tomlandia:
OK just removed a forgotten debugger statement and merged

tomlandia:
After-merge fix: `src/core/javascripts/services/i18n/en-gb.js` was one subdirectory too much to be compiled so I renamed it to `src/core/javascripts/components/moment-en-gb.js` in a10daff

PR #302, opened 2016-06-08
[COS-30] Added functions to manipulate multiple spaces in complex tic…

gmunumel:
https://bookingbug.atlassian.net/browse/COS-30


src/core/javascripts/models/event.js.coffee
```diff
@@ -238,3 +238,34 @@ angular.module('BB.Models').factory "EventModel", ($q, BBModel, BaseModel, DateT
...
+    * 
+    *
+    * @returns {bolean} true is all the quatities are bigger than zero
+    ###
+    allTicketsQtyBiggerThanZero: () ->
+      for ticket in @tickets
+        if ticket.qty <= 0
+          return false
+      return true
+
```
lukeferguson: @gmunumel since the methods essentially are the opposite of each other, I would just keep this as on method called `hasTicketsSelected` which you can use the boolean result to disable/enable buttons.

src/core/javascripts/models/event_ticket.js.coffee
```diff
@@ -70,6 +70,29 @@ angular.module('BB.Models').factory "EventTicketModel", ($q, BBModel, BaseModel)
...
+          @max_left = event.ticket_spaces[pool_id].left
+
+      if @max_left == 0
+        [0]
+      else
+        [0].concat [@min..@max_left]
+
+
+    ###**
+    * @ngdoc method
```
lukeferguson: If using complex tickets, and there's availability in at least one of the pools, use rangeTicketSpaces else use getRanges to calc range.  rangeTicketSpaces should also only be used when what's left in the pool is less than `ticket.max_num_bookings`.

src/core/javascripts/models/event_ticket.js.coffee
```diff
@@ -70,6 +70,47 @@ angular.module('BB.Models').factory "EventTicketModel", ($q, BBModel, BaseModel)
...
+    * @name getRangeTicketSpaces
+    * @methodOf BB.Models:EventTicket
+    * @description
+    * Get a range base on the minimum number of bookings and the tickets left. This method
+    * is useful when you have multiple spaces for an event in complex tickets. 
+    *
+    * @returns {array} The returned range
+    ###
+    getRangeTicketSpaces: (pool_id, event) ->
+      @max_left = 0
```
lukeferguson: @gmunumel `@` and `this` are the same - there's no need to assign this value to object though just keep it as a local one

src/core/javascripts/models/event_ticket.js.coffee
```diff
@@ -70,6 +70,47 @@ angular.module('BB.Models').factory "EventTicketModel", ($q, BBModel, BaseModel)
...
+      @min = @min_num_bookings
+      if this.isAvailable(event)
```
lukeferguson: favour `@` over `this`

src/core/javascripts/models/event_ticket.js.coffee
```diff
@@ -70,6 +70,47 @@ angular.module('BB.Models').factory "EventTicketModel", ($q, BBModel, BaseModel)
...
+    * @methodOf BB.Models:EventTicket
+    * @description
+    * Get a range base on the minimum number of bookings and the tickets left. This method
+    * is useful when you have multiple spaces for an event in complex tickets. 
+    *
+    * @returns {array} The returned range
+    ###
+    getRangeTicketSpaces: (pool_id, event) ->
+      @max_left = 0
+      @min = @min_num_bookings
```
lukeferguson: same again gain, no need for this to be an instance variable

PR #307, opened 2016-06-10
[BW-38/57] Use 1.2.28-compatible ngAria 1.5 fork to increase accessibility

tomlandia:


lukeferguson:
@tomlandia you can now add this is an a bower dependency rather than manually as we've now migrated to angular 1.5

tomlandia:
Closed as `angular` and `angular-aria` v1.5.8 are now bower dependencies of `bookingbug-angular-core`.

PR #310, opened 2016-06-15
[ANZ-23] fix for scrollbars trigering select event

lefterisk:


edclements:
It would be nice to fix upstream.

PR #316, opened 2016-06-21
Update member-form directive to allow use of custom template. If memb…

WesleyPumpkinhead:
…er_schema is used (no custom template) then override answers in Client.questions with those in the Client.q objects


lukeferguson:
Looks good @WesleyPumpkinhead. Did ou forget to commit the ` _member_form`? template?

PR #317, opened 2016-06-22
Gulp fix for mainBowerFiles module.

victorjeman:
Gulp fix. When using mainBowerFiles for css some @imports rules where added to final concatenated file.


victorjeman:
some @import rules where copied to the final concatenated css file

PR #318, opened 2016-06-22
Changes to fix mobile number field from clearing and "failing" valida…

Joensg:
…tion on form submission


Joensg:
@lukeferguson Please see also PR #180 in bookingbug-angular-bespoke

src/core/javascripts/directives/helpers.js.coffee
```diff
@@ -181,17 +181,18 @@ app.directive 'bbDebounce', ($timeout) ->
...
   restrict: 'A',
   require: 'ngModel',
   link: (scope, element, attrs, ctrl) ->
 
     prettyifyNumber = (value) ->
-      if value and value[0] != "0"
-        value = "0" + value
+      if scope.client_mobile
+        value = scope.client_mobile
+      else if value and value[0] != "0"
```
lukeferguson: Checking for for zero infers that all local numbers will start with zero when that isn't nessecarily the case.  The problem is just removing it might lead to number getting double formatted... can you test? We might need to write a parser that then strips the zero off.

Joensg: I did test it. The number could be of two forms, 07123456789, +447123456789.
If you enter 7123456789(without leading zero or +), it fails validation.
As to the view on the front-end, I used the ng-change to maintain the same value as entered by the user. The server returns the the number 7123456789 with prefix 44 in a property whether we enter the number as  07123456789 or  +447123456789.

PR #319, opened 2016-06-23
Dun 45

DeanPalmer:
Check if there is no availability in the initial time range and offer the client to jump to the first available day.


src/core/javascripts/services/viewport_size.js.coffee
```diff
@@ -0,0 +1,22 @@
...
+	viewport_size = null
+
+	setViewportSize: (size) ->
+		if size != viewport_size
+			viewport_size = size
+			$rootScope.$broadcast 'ViewportSize:changed'
+
+	getViewportSize: () ->
+		return viewport_size
+]
```
lukeferguson: indentation should use spaces and be tabs should 2 spaces @DeanPalmer

src/public-booking/templates/_week_calendar.html
```diff
@@ -1,4 +1,8 @@
 <div>
+
+  <div class="alert alert-info cursor-pointer" ng-show="no_slots_in_week" ng-click="showFirstAvailableDay()"><i class="fa fa-info-circle" aria-hidden="true"></i> Looks like there's no availability for the current time range, click to jump to first available day.
```
lukeferguson: reword to "It looks like there's no availability in the next <time range length> days.  

then add button that says (aligned to the right, vertically aligned)
"Jump to Next Available Day""

Remember that user don;t click on mobile. Button infers that it needs to be clicked too - a visual affordance in design speak :)

PR #321, opened 2016-06-23
Updated bb-durations to be more dynamic and rebuild durations array w…

WesleyPumpkinhead:
…henever service is changed in scope


WesleyPumpkinhead:
@lukeferguson Small change to make bb-durations more dynamic

PR #324, opened 2016-06-27
ANZ-66 avoid using select slot while loading data for currentItemUpdate

fnicola:


src/core/javascripts/controllers/time_list.js.coffee
```diff
@@ -246,27 +247,29 @@ angular.module('BB.Controllers').controller 'TimeList', ($attrs, $element, $scop
...
-        else if requested_slot is null
-          $scope.availability_conflict = true
+        slot_checks(time_slots) if options.check_requested_slot == true
 
       , (err) -> $scope.setLoadedAndShowError($scope, err, 'Sorry, something went wrong')
 
     else
       $scope.setLoaded $scope
 
+  slot_checks = (time_slots) ->
```
lukeferguson: Use camelCase for method name.  Rename to checkRequestedSlot - nice to use a lead method names with a verb to describe what they do

PR #325, opened 2016-06-28
Crated a directive for file upload functionality.

victorjeman:


src/core/templates/file_upload.html
```diff
@@ -0,0 +1,45 @@
+<label for="attendee_last_name"
```
lukeferguson: `attendee_last_name` is incorrect

src/core/templates/file_upload.html
```diff
@@ -0,0 +1,45 @@
...
+  class="control-label col-sm-4">
+    File upload:
+</label>
+
+<div class="col-sm-5 ng-scope">
```
lukeferguson: No need to reference `ng-scope`

PR #327, opened 2016-06-28
fix for calendar breaking on ee-digital

Joensg:
Also please see https://github.com/BookingBug/bookingbug-angular-bespoke/pull/199


src/core/javascripts/directives/helpers.js.coffee
```diff
@@ -181,18 +181,28 @@ app.directive 'bbDebounce', ($timeout) ->
...
 # Adds a formatter that prepends the model value with zero. This is useful for
 # nicely formatting numbers where the prefix has been stripped, i.e. '7875123456'
-app.directive 'bbLocalNumber', () ->
+# Adds a parser to store user entered value and use this in the formatter
+app.directive 'bbLocalNumber', ($filter) ->
   restrict: 'A',
   require: 'ngModel',
   link: (scope, element, attrs, ctrl) ->
 
+    userinput_mobile = scope.client.mobile
```
lukeferguson: hey @Joensg, it's bad to reference a named scope variable as this might change depending which model the directive is bound to.  the ngModel controller allows you to reretireve the view value like so: `ctrl.$viewValue`.  Please update and retest

Joensg: I just needed the userinput_mobile to be defined and set to something. Otherwise the parser was giving an error. It works if I set it to zero.

Joensg:
## Screencast with "+44" prefix:
[https://gyazo.com/60d721f9f9e49a48060518938ee29b10](https://gyazo.com/60d721f9f9e49a48060518938ee29b10)

## Screencast with "0" prefix:
[https://gyazo.com/65b7e21dda437754b1fd46f94c0ffd77](https://gyazo.com/65b7e21dda437754b1fd46f94c0ffd77)

src/core/javascripts/directives/helpers.js.coffee
```diff
@@ -181,18 +181,28 @@ app.directive 'bbDebounce', ($timeout) ->
...
+      if value
+        userinput_mobile = value
+      return value
+
     prettyifyNumber = (value) ->
-      if value and value[0] != "0"
-        value = "0" + value
+      if userinput_mobile
+        value = userinput_mobile
+      else if value and value[0] != "0"
```
lukeferguson: We shouldn;'t really be checking for for "0" as that's only for local UK numbers. can you test what effect just changing this to an else would have?

Joensg: @lukeferguson it works the same. Removed the zero. Changes in new commit below.

src/core/javascripts/directives/helpers.js.coffee
```diff
@@ -181,18 +181,26 @@ app.directive 'bbDebounce', ($timeout) ->
...
 # Adds a formatter that prepends the model value with zero. This is useful for
 # nicely formatting numbers where the prefix has been stripped, i.e. '7875123456'
-app.directive 'bbLocalNumber', () ->
+# Adds a parser to store user entered value and use this in the formatter
+app.directive 'bbLocalNumber', ($filter) ->
   restrict: 'A',
   require: 'ngModel',
   link: (scope, element, attrs, ctrl) ->
 
+    userinput_mobile = null
```
lukeferguson: Consider making this a scoped var

Joensg: @lukeferguson Changed userinput_mobile to a scope variable and isolated the scope. Please see new commit below. Thanks! :)

PR #328, opened 2016-06-28
[COS-8] File upload directive

victorjeman:


lukeferguson:
Looks good @victorjeman, given the fact that you're removing the onFileSelect method from teh item details and purchase directives, this will be a breaking change.  Can you please write something to add to the "Breaking Changes" page in the wiki discussing how to migrate to your new directive

PR #334, opened 2016-07-01
[ANZ-83] Fixing timezone issue in calendar datepicker

lefterisk:


edclements:
Can this be encapsulated better by including it in the relevant directive?

PR #336, opened 2016-07-05
added in a check for item_default being defined in method checkDefaul…

Joensg:
…tTime


Joensg:
## Screencast before fix:


## Screencast after applying fix:

lukeferguson:
@Joensg: The `item_defaults` are not passed in the `checkDefaultTime` time method when they should be

PR #353, opened 2016-07-13
[COS-38] file upload validation

victorjeman:
@lukeferguson I've made a clean PR for a better git  history. If this is ok I should make a little change for COS project and I should know what type of files should be accepted there, and what should be the file size limit.


lukeferguson:
Looks good @victorjeman. Go ahead and update the COS project.

PR #354, opened 2016-07-13
[BW-111][CISCO-56][CISCO-68] Timezone fixes for view/cancel/reschedule journey

tomlandia:
1. `DatetimeUtilities.convertTimeSlotToMoment` now accepts both SDK `day` objects and standard `moment` objects as both are being passed to it (CISCO-68).
2. If `use_local_timezone` is enabled, the child company timezone is now set at purchase page to display times with the right timezone on this page (CISCO-56).


src/core/javascripts/purchase/controllers/purchase.js.coffee
```diff
@@ -18,6 +18,7 @@ angular.module('BB.Controllers').controller 'Purchase', ($scope,  $rootScope, Co
     $scope.bb.company = new BBModel.Company(company)
     $scope.company = $scope.bb.company
     $scope.bb.item_defaults.company = $scope.bb.company
+    SettingsService.setDisplayTimeZone(company.timezone) if SettingsService.getUseCompanyTimeZone()
```
lukeferguson: Was this tested @tomlandia? `getUseCompanyTimeZone` is not a method in the SDK?

src/core/javascripts/services/datetime_utilities.coffee
```diff
@@ -2,8 +2,8 @@ angular.module('BB.Services').factory "DateTimeUtilitiesService", (SettingsServi
 
   # converts date and time belonging to BBModel.Day and BBModel.TimeSlot into
   # a valid moment object
-  convertTimeSlotToMoment: (day, time_slot) ->
-    return if !day and !time_slot
+  convertTimeSlotToMoment: (date, time_slot) ->
```
lukeferguson: In almost all instances where `convertTimeSlotToMoment` is called, a day object is passed in @tomlandia.  Whilst I am happy with the change to be made given that the date component is the only thing used from the day object, I expect more rigorous testing and verification to be completed before a PR is raised.

lukeferguson:
Reviewed. Twe comments @tomlandia.

src/core/javascripts/services/datetime_utilities.coffee
lukeferguson: Given the date component from the Day object is the only thing used @tomlandia, I think it would be cleaner if you just update the SDK to always pass that in.  You can them make the method return if the `date` is not a moment object.

tomlandia:
Sorry this line is useless and does crash the function. I missed it because it didn't affect that particular journey.

PR #356, opened 2016-07-13
[SKYG-108] left/right calendar arrows black/grey bug fix

victorjeman:


victorjeman:
skygarden has a  custom  _event_lists.scss in src. Changes should be made there for that project.

PR #366, opened 2016-07-15
provide an option to set the default calendar view to single-day, so …

Joensg:
…that it speeds up loading time especially for anz-appointments project when loading the 'Add Appointment' journey route.


Joensg:
Changes that enable providing an option to support loading single-day view as default the view for the calendar page. This is especially as in the case of anz-appointments project for the 'Add Appointment' journey route.

Joensg:
Please also see https://github.com/BookingBug/bookingbug-angular-bespoke/pull/209

PR #378, opened 2016-07-26
event_list template fix identation and others small html fixes

victorjeman:
PR for feedback purposes!
Hi @lukeferguson, while I was waiting for feedback on other PR I made some small changes in on of the templates.

I was thinking that we could clean out a little all the templates. We could sort out the indentation bugs and other small issues line line styles.

We can denote thematic breaks or 'big blocks' in content with five (4) empty lines.
template

Separate independent but loosely related snippets of markup with a single empty line.

We could put the element attributes in a new line if they are passing the 80 characters limit.
new-line

I don't think there is a need for a fancy guideline, just a couple of things that we should be consistent about through all the templates could increase the markup readability.

### DIFF
--- a/src/public-booking/templates/event_list.html
+++ b/src/public-booking/templates/event_list.html
@@ -1,8 +1,9 @@
-<div bb-events="{mode: 2}" ng-init="bb.basket.clear()">
+<!-- bb-events -->
+<section bb-events="{mode: 2}" ng-init="bb.basket.clear()">
 
-  <div class="page-header">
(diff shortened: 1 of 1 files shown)


lukeferguson:
Hey @victorjeman - I like all of this - feel free update the styleguide with this: https://docs.google.com/document/d/10ObqRqngf-SFXLdLpm8RRfok8F7GeIsiJh0PJ8wY0Nk/edit?usp=sharing

PR #395, opened 2016-08-02
en-gb locale conditions extended to Australian clients as they are similar

gwenau:
Australian locale conditions are more similar to the British locale conditions than the US locale conditions so the en-gb locale conditions are extended to them.


lukeferguson:
As discussed, utilise the en-au locale instead @gwenau

WesleyPumpkinhead:
@lukeferguson I have refactored this to use moment.locale when the country_code is "gb" or "au" by using RegExp and removed what I consider to be a pointless "country_code = value" line (although this does appear to be a convention for the other functions in settings.js.coffee)

PR #396, opened 2016-08-02
[COS-38] Update for file upload file validation

victorjeman:
Hi @lukeferguson , I've made the changes for file upload but it seems too much, couldn't we use another attribute with pretty file types and then just expose that to user when the selected file is wrong.

something like: 

screen shot 2016-08-02 at 12 57 45

If the way I did it is ok then I should change in COS project the accept attribute from
screen shot 2016-08-02 at 13 04 06
to
screen shot 2016-08-02 at 13 03 34
because we are using **ngf-accept** instead of **accept**


lukeferguson:
Hey @victorjeman - yep go with your suggested approach :)

victorjeman:
Changes are done!

PR #402, opened 2016-08-03
[MHM-20] Ability to pass in the long_id into the widget during initialisation.

Joensg:
Added in long_id to item defaults so that it can be passed into the widget under item_defaults during widget initialisation for amend/cancel journey.

## Screencast with these changes implemented:
_(Note how no id is being passed in the url)_


src/core/javascripts/controllers/base.js.coffee
```diff
@@ -544,7 +544,7 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location,
           $scope.bb.admin = admin
         setup_promises.push sso_admin_login
 
-      total_id = QueryStringService('total_id')
+      total_id = QueryStringService('total_id') || (if $scope.bb.item_defaults? then $scope.bb.item_defaults.long_id)
```
lukeferguson: Hey @Joensg - this is bit hard to read with an if statement inside `||` assignment. I'd split it out over two lines to improve readability.

Joensg:
Regarding the commit for the bug fix, notice how the id was being passed into $scope.load() at https://github.com/BookingBug/bookingbug-angular/pull/402/files#diff-870743f9d4ee491bc0f411cac58fdd8eR47 but was being overridden later.

PR #406, opened 2016-08-05
Service List Optimisations 

Morphasis:
**Reason for change:** The service list never worked well when the other settings were enabled. As we are moving to a more out of the box solution i thought it might be a idea to make this template actaully functional.

**Breaking changes:**
None it supports all the old features just works better.

**Things to note:**
n/a as its a template change. 

Before hand for reference:
Screenshot:
https://i.gyazo.com/4ccb73afc19c1bff8d092565567be993.png

Works well with mixed settings:
screenshot:
https://i.gyazo.com/f69c051a1aecd18d9108e6859e5b6def.gif
Can have prices without breaking formatting
screenshot:
https://i.gyazo.com/384cae2af0d885493dd541963e5812b1.png
The images are now too supported without it breaking
screenshot:
https://i.gyazo.com/0e6ca856b99e6c6f2e6cb48e2bf9454e.png

### DIFF
--- a/src/core/javascripts/services/viewport_size.js.coffee
+++ b/src/core/javascripts/services/viewport_size.js.coffee
@@ -5,18 +5,30 @@
 * @name BB.Services.service:ViewportSize
 *
 * @description
-* Stores the current screen size breakpoint. 
+* Stores the current screen size breakpoint.
 ###
 angular.module('BB.Services').factory 'ViewportSize', ['$rootScope',($rootScope) ->
 
   viewport_size = null
 
   setViewportSize: (size) ->
-    if size != viewport_size
+    # console.log viewport_size, "viewport_size"
+    # console.log size, "size"
+    num = 1 if viewport_size == 'xs'
+    num = 3 if viewport_size == 'sm'
+    num = 5 if viewport_size == 'md'
+    num = 7 if viewport_size == 'lg'
+    num2 = 1 if size == 'xs'
+    num2 = 3 if size == 'sm'
+    num2 = 5 if size == 'md'
+    num2 = 7 if size == 'lg'
+    console.log num2
+    console.log num
+    if num < num2 or !num
+      console.log "asdasdjkaskhjdaskhj"
       viewport_size = size
       $rootScope.$broadcast 'ViewportSize:changed'
 
   getViewportSize: () ->
     return viewport_size
 ]
-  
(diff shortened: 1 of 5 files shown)


DeanPalmer:
Border bottom for last child may be on the wrong element, all panels have bottom borders causing styling issues. Tested in project with no custom css.

lukeferguson:
@Morphasis little more refinement required i think - display of the price/image shouldn't mess with the line spacing between the service name and description.

I think it would nicer if the image was a little thumbnail in a column on the left most edge when present - we also want to constrain the image size rather than  rendering it full size too.

Morphasis:
Hi there guys i did not really have time to test this properly with the remaining changes. So here are the change log.
screenshots:
https://gyazo.com/a1bd83519452a2f4779268d30d31265b
https://gyazo.com/f46009a90d7ebdc6595626929776f3d4
https://gyazo.com/afb6cf13ca9423c4895afbeeb0cd1bce

- Fixed repeat dean bug.
- Polished styling tested all use cases. (which i did not have time before)

Apologies for the poor PR i was very busy when i made this. Ill make sure they are better from now on.

src/core/javascripts/services/viewport_size.js.coffee
```diff
@@ -5,18 +5,30 @@
...
   viewport_size = null
 
   setViewportSize: (size) ->
-    if size != viewport_size
+    # console.log viewport_size, "viewport_size"
+    # console.log size, "size"
+    num = 1 if viewport_size == 'xs'
+    num = 3 if viewport_size == 'sm'
+    num = 5 if viewport_size == 'md'
+    num = 7 if viewport_size == 'lg'
```
lukeferguson: @Morphasis these numbers seem to reflect the various different time range list - this breaks the separation of concerns between the ViewportSize service and the time range list directive.

src/core/javascripts/services/viewport_size.js.coffee
```diff
@@ -5,18 +5,30 @@
...
+    num2 = 1 if size == 'xs'
+    num2 = 3 if size == 'sm'
+    num2 = 5 if size == 'md'
+    num2 = 7 if size == 'lg'
+    console.log num2
+    console.log num
+    if num < num2 or !num
+      console.log "asdasdjkaskhjdaskhj"
```
lukeferguson: Remove log

lukeferguson:
Service list has been updated independently of this PR

PR #407, opened 2016-08-05
add error css styling to datepicker

MarkShakespeare:


src/public-booking/stylesheets/_form.scss
```diff
@@ -118,3 +118,9 @@
   }
 
 }
+
+.has-error .btn-datepicker {
```
lukeferguson: selector should be `.has-error .input-group-btn .btn` so that it works for for all button addons rather than just the date picker.  if .btn doest work, target `btn-default ` instead

MarkShakespeare:
updated it

PR #408, opened 2016-08-05
update bb-payment to get absolute url from  service instead, so that …

WesleyPumpkinhead:
…the custom_stylesheet property is set to the name of the custom stylesheet only (and not the url). This change should be backwards compatible.


src/core/javascripts/controllers/payment.js.coffee
```diff
@@ -39,7 +39,15 @@ angular.module('BB.Directives').directive 'bbPayment', ($window, $location, $sce
       referrer = $location.protocol() + "://" + $location.host()
       if $location.port()
         referrer += ":" + $location.port()
-      custom_stylesheet = scope.payment_options.custom_stylesheet if scope.payment_options.custom_stylesheet
+
+      if scope.payment_options.custom_stylesheet
+        if custom_stylesheet.match(/http/)
+          # custom stylesheet as an absolute url, for ex. "http://bespoke.bookingbug.com/staging/custom-booking-widget.css"
+          custom_stylesheet = scope.payment_options.custom_stylesheet if scope.payment_options.custom_stylesheet
```
lukeferguson: No need for `if scope.payment_options.custom_stylesheet` given you've already checked that at line 43

WesleyPumpkinhead:
@lukeferguson removed unnecessary if condition as per your comment

PR #414, opened 2016-08-11
add option to pay from upcoming bookings if you have only reserved a …

MarkShakespeare:
…space


src/member/javascripts/controllers/member_bookings.js.coffee
```diff
@@ -160,10 +160,10 @@ angular.module('BBMember').controller 'MemberBookings', ($scope, $modal, $log, M
 
   book: (booking) ->
    
-    $scope.notLoaded $scope
+    $scope.notLoaded $scope
```
lukeferguson: @MarkShakespeare remove leading spaces - check your editor settings to make sure it's not adding them unintentionally too

PR #420, opened 2016-08-12
Wait 49

WesleyPumpkinhead:


src/public-booking/stylesheets/_fonts.scss
```diff
@@ -66,8 +66,16 @@
 
 .icon-price {
   font-family: 'Glyphicons Halflings';
+  vertical-align: middle;
   &:before {
     content: "\e149";
   }
 }
 
+.icon-info-sign{
```
lukeferguson: We shouldn't really be adding icons when they're Glyphicons already?

src/public-booking/templates/event_list.html
```diff
@@ -219,8 +219,24 @@ <h3 class="truncate-sm">{{item.description}}</h3>
...
+                      <span ng-show='isCollapsed'>Less Info</span>
+                    </button>
                   </div>
                 </div>
+                 <div class="row" ng-show="item.group.description">
+                   <div class="col-sm-10 col-xs-12 event-details">
+                     <ul class="event-summary-list" collapse="!isCollapsed">
+                       <li>
+                         <p class='event-group-description'>
+                           <span class="bb-icon icon-info-sign"></span>
```
lukeferguson: Why didn't you just use a glyph icon or a font awesome one?

src/public-booking/stylesheets/_fonts.scss
```diff
@@ -66,8 +66,16 @@
 
 .icon-price {
   font-family: 'Glyphicons Halflings';
+  vertical-align: middle;
```
lukeferguson: Maybe this should be removed too if it's just a glyhicon?

WesleyPumpkinhead: I only added the vertical-align: middle because I noticed this particular icon was quite badly aligned, but I will remove this.

src/public-booking/templates/event_list.html
WesleyPumpkinhead: I think it was something to do with the styling rules of bb-icon - the first thing I would have done is just to use fa fa-xxx but there was some issues as I remember, so I decided to add them into the bb-icon class instead. But I really don't remember what the issue was, and people can fix that locally in their projects anyway, so I am just going to change this to FontAwesome class as you suggest.

WesleyPumpkinhead:
@lukeferguson updated as per your comments

PR #421, opened 2016-08-15
MemberPurchaseService Improvements

Joensg:
Changed 'deferred.reject(err)' to 'deferred.resolve.([])' to take care of the case when there are no purchases and the server returns with a '404', thus leading to '$scope.purchases' to be undefined.

<h3 align="center">
Error before applying the fix :
</h3>

=============

<h3 align="center">
Error seen when using 'deferred.reject([])' instead of 'deferred.resolve([])' :
</h3>

=============

<h3 align="center">
After applying the fix with 'deferred.resolve([])' :
</h3>

=============


Joensg:
<h3 align="center">
Error before applying the fix :
</h3>

src/member/javascripts/services/purchase.js.coffee
```diff
@@ -14,7 +14,7 @@ angular.module('BBMember.Services').factory "MemberPurchaseService", ($q, $rootS
             new BBModel.PurchaseTotal(purchase)
           deferred.resolve(purchases)
         , (err) ->
-          deferred.reject(err)
+          deferred.resolve([])
```
lukeferguson: Does the errr object come back with the 404 status? We only want to resolve the promise in this scenario as we might hide genuine errors with this approach.

PR #422, opened 2016-08-15
Member purchase template change

Joensg:
For commit : https://github.com/BookingBug/bookingbug-angular/commit/4221724e136a90c03910a9959515d2d132f2c521
=============

<h3 align="center">
Before:  adding padding-top of '5px' :
</h3>

<h3 align="center">
After:  adding padding-top of '5px' :
</h3>

For commit : https://github.com/BookingBug/bookingbug-angular/commit/caa1a26491e0dec3a89ba04a7f25aa80d3550867
=============

<h3 align="center">Before: fixing broken ng-show='purchases.length == 0' :</h3>

<h3 align="center">After: fixing broken ng-show='purchases.length == 0' :</h3>


src/member/templates/member_purchases.html
```diff
@@ -2,11 +2,11 @@
 
   <div>
 
-    <div ng-show="purchases.length == 0">
+    <div ng-show="!purchases || purchases.length == 0">
```
lukeferguson: `!purchases` will evaluate as falsey immediately as the page loads - we do want to show any message until we have a response from the API

PR #425, opened 2016-08-16
[IMPL-43] 

TolgaCura:
- Lazy loading events when max pages to display is reached
- Filter events now apply to 300 events rather than 100
- Pagination summary reflects information based on number of items per page and max pages to display


edclements:
Is this using the API pagination correctly?

TolgaCura:
original issue marked as resolved and no longer an issue

PR #426, opened 2016-08-17
Handle display timezone different from system timezone

tomlandia:
- Adjust to company timezone even if displayTimeZone == company time zone as not adjusting causes shifts if you're adjusting to company time zone from another time zone.
- Don't adjust calendar headers one day back if the display time zone is west of the system.

### DIFF
--- a/src/public-booking/templates/_week_calendar.html
+++ b/src/public-booking/templates/_week_calendar.html
@@ -127,8 +127,8 @@
             </button>
           </span>
           <h3 class="day-heading col-xs-8 col-sm-12">
-            <span class="visible-xs">{{day.date | datetime: "ddd"}} <strong>{{day.date | datetime: "D"}} </strong>{{day.date | datetime: "MMM"}}</span>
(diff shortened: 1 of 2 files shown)


src/public-booking/templates/_week_calendar.html
```diff
@@ -127,8 +127,8 @@ <h2 class="month-heading">{{pretty_month_title('MMMM', 'YYYY')}}</h2>
             </button>
           </span>
           <h3 class="day-heading col-xs-8 col-sm-12">
-            <span class="visible-xs">{{day.date | datetime: "ddd"}} <strong>{{day.date | datetime: "D"}} </strong>{{day.date | datetime: "MMM"}}</span>
-            <span class="hidden-xs">{{day.date | datetime: "ddd"}} <strong>{{day.date | datetime: "D"}} </strong></span>
+            <span class="visible-xs">{{day.date.format("ddd")} <strong>{{day.date.format("D")} </strong>{{day.date.format("MMM")}</span>
```
lukeferguson: Should use datetiem from first time slot (when available) to ensure date header shifts correctly as slots switch days. If no slots are available, use `format` as proposed.

damianpudlo:
obsoleted

PR #427, opened 2016-08-17
PR-unwrap-DRY

Morphasis:
There are three repeated methods in this moved them to pass in those parameters and then used those methods across the file.


Morphasis:
Im going to run through some test cases and then post the findings so please do not merge before hand. Because its quite a large change.

Morphasis:
will not work with v2

PR #429, opened 2016-08-18
Autofill support

MarkShakespeare:


src/public-booking/templates/event.html
```diff
@@ -223,7 +223,7 @@ <h2 ng-show="!bb.basket.hasWaitlistItem()" class="heading" ng-bind="!selected_ti
             <div class="form-group" ng-class="{'has-error': client_form.address1.$invalid && ((client_form.address1.$dirty && !client_form.address1.$focused) || booking_form.submitted)}">
               <label for="address1" class="control-label col-sm-4">Address:</label>
               <div class="col-sm-5">
-                <input type="text" name="address1" id="address1" ng-model="client.address1" class="form-control"/>
+                <input type="text" name="address1" id="address1" ng-model="client.address1" class="form-control" autocomplete="street-address"/>
```
lukeferguson: [Docs](https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill?hl=en) for two inputs that you should use address-line1 and address-line2

PR #430, opened 2016-08-18
Pr unwrap dry v1

Morphasis:
This is the same pull request for v2 but against the v1 branch.

This updates the unwrap file so that it reuses the code that it repeated all of the time into one consistent reusable service. Therefore being shorter more readable and more maintainable.


src/core/javascripts/services/unwrap.js.coffee
```diff
@@ -1,238 +1,138 @@
...
     if angular.isArray(resource)
 
-      models = (new BBModel.Service(service) for service in resource)
+      models = (new model(service) for service in resource)
       deferred.resolve(models)
 
-    else
-    
-      resource.$get('services').then (items) =>
+    else if resource.$has(key)
```
lukeferguson: Add an else statement to reject the promise

PR #435, opened 2016-08-24
Renamed 'long_id' to 'purchase_total_long_id'

Joensg:
Renamed '$scope.bb.item_defaults.long_id' to '$scope.bb.item_defaults.purchase_total_id'.


Glenn:
I'm not sure I like this!
This will break all existing uses. PLus it's not the database id - which is what I would expect this to refer to. Why are we doing this ???

lukeferguson:
The change to add `long_id` was only introduced 19 days ago @Glenn - poor naming that slipped through my initial review - item defaults should be self explanatory for what they represent.

Glenn:
fair enough ;-)

lukeferguson:
I do agree with @Glenn that `purchase_total_id` could be misleading @Joensg - `purchase_total_long_id` would be best!

Joensg:
@lukeferguson ok changed as requested. :)

src/core/javascripts/controllers/base.js.coffee
```diff
@@ -545,8 +545,8 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location, $rootS
           $scope.bb.admin = admin
         setup_promises.push sso_admin_login
 
-      if $scope.bb.item_defaults and $scope.bb.item_defaults.long_id
-        total_id = $scope.bb.item_defaults.long_id
+      if $scope.bb.item_defaults and $scope.bb.item_defaults.purchase_total_long_id
```
WesleyPumpkinhead: @Joensg it should be purchase_total_id and not purchase_total_long_id

src/core/javascripts/controllers/base.js.coffee
```diff
@@ -545,8 +545,8 @@ angular.module('BB.Controllers').controller 'BBCtrl', ($scope, $location, $rootS
...
+        total_id = $scope.bb.item_defaults.purchase_total_long_id
```
WesleyPumpkinhead: @Joensg see above comment

PR #439, opened 2016-08-31
BW-110 Localized datetime format tokens updated

victorjeman:


victorjeman:
>"h:mma" => "LT"
>"time" => "LT"
>"h[:]mma" => "LT"
>`HH:mm` => "LT"
> `time`=> `LT`

> 'date` => "L"

>`Do MMM YYYY HH:mm` and `Do MMM YYYY h.mma` => " LLL"

>`DD/MM/YYYY HH:mm` should be split into two filters using "L" and "LT" respectively

PR #444, opened 2016-09-02
translations

damianpudlo:
…o 'sanitizeParameters'


lukeferguson:
@damianpudlo angular translate docs suggest favouring the translate directive over the filter as filter creates too many watch expressions.

See https://angular-translate.github.io/docs/#/api/pascalprecht.translate.$translate

src/admin/templates/cancel_booking_modal_form.html
```diff
@@ -8,17 +8,17 @@ <h3 class="modal-title">Cancel Booking</h3>
     <div class="checkbox">
       <label>
         <input type="checkbox" name="notify" ng-model="model.notify">
-        Send cancellation confirmation to {{booking.client_email}}?
+        <span translate translate-values="{email:'{{booking.client_email}}'}">ADMIN.MODAL.CANCEL_BOOKING.SEND_EMAIL</span>
```
lukeferguson: @damianpudlo docs suggest you don;t need to wrap `booking.client_email` as an expression.

src/admin/templates/cancel_booking_modal_form.html
```diff
@@ -1,5 +1,5 @@
 <div class="modal-header">
-  <h3 class="modal-title">Cancel Booking</h3>
+  <h3 class="modal-title"><span translate>ADMIN.MODAL.CANCEL_BOOKING.TITLE</span></h3>
```
lukeferguson: Use `<span translate="ADMIN.MODAL.CANCEL_BOOKING.TITLE"></span>` so it's consistent with public booking and the way `translate-values` are expressed.

PR #448, opened 2016-09-09
Month picker now uses angular carousel

TolgaCura:


src/core/javascripts/directives/month_picker.js.coffee
```diff
@@ -12,14 +12,24 @@ angular.module('BB.Directives').directive 'bbMonthPicker', (PathSvc, $timeout) -
     scope.picker_settings = scope.$eval(attrs.bbMonthPicker) or {}
     scope.picker_settings.months_to_show = scope.picker_settings.months_to_show or 3
 
-    stopWatch = scope.$watch attrs.dayData, (dates) ->
-      if dates
-        scope.processDates(dates)
-        stopWatch()
+    $(window).resize () ->
+      scope.$apply () ->
```
lukeferguson: Might want to consider putting this under a timeout to increase performance

src/core/javascripts/directives/bb-month-picker-listener.js.coffee
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+angular.module('BB.Directives').directive 'bbMonthPickerListener', (PathSvc, $timeout) ->
```
lukeferguson: Does the listener need to be a separate directive?

src/core/javascripts/directives/month_picker.js.coffee
```diff
@@ -137,6 +128,89 @@ angular.module('BB.Directives').directive 'bbMonthPicker', (PathSvc, $timeout) -
...
+        date = cur_month.clone().startOf('week')
+        month = {weeks: []}
+        month.index = m - 1
+        for w in [1..6]
+          week = {days: []}
+          for d in [1..7]
+
+            month.start_date = date.clone() if date.isSame(date.clone().startOf('month'),'day') and !month.start_date
+
+            day = {
```
lukeferguson: Is this loop not rpeating some of the code in `processDates` ?

src/public-booking/bower.json
```diff
@@ -23,6 +23,7 @@
   "dependencies": {
     "bookingbug-angular-core": "master",
     "iframe-resizer": "~3.5.5",
-    "animate.css": "~3.5.2"
+    "animate.css": "~3.5.2",
+    "angular-carousel": "~0.3.13"
```
lukeferguson: Slick needs to removed from the core bower.json and main.js.coffee in core too

src/core/javascripts/directives/bb-month-picker-listener.js.coffee
TolgaCura: I've played around with having it in the month picker directive, however the carousel doesn't seem to play nice. The built in indicators do not match what is being displayed, for example it hides the left indicator when the carousel isn't yet displaying the first card.

src/core/javascripts/directives/month_picker.js.coffee
```diff
@@ -13,7 +13,7 @@ angular.module('BB.Directives').directive 'bbMonthPicker', (PathSvc, $timeout) -
     scope.picker_settings.months_to_show = scope.picker_settings.months_to_show or 3
 
     $(window).resize () ->
-      scope.$apply () ->
+      $timeout ->
```
lukeferguson: Set the timeout to be 500ms or maybe a bit less, it defaults to 0 i think

src/core/javascripts/directives/month_picker.js.coffee
```diff
@@ -144,33 +110,50 @@ angular.module('BB.Directives').directive 'bbMonthPicker', (PathSvc, $timeout) -
         $scope.months = months if months.length
         # set required vars
         cur_month = angular.copy($scope.months[$scope.months.length - 1].start_date)
+        cur_month.add(1, 'month')
```
lukeferguson: Hey @TolgaCura - `cur_month` is a bit misleading as it's actually the last month, or rather then end of the month range we're trying to set.

src/core/javascripts/directives/month_picker.js.coffee
```diff
@@ -144,33 +110,50 @@ angular.module('BB.Directives').directive 'bbMonthPicker', (PathSvc, $timeout) -
...
 
       monthCollection = []
       slide = []
       
       $scope.months = [] if !$scope.months
 
-      for m in [0...num_empty_months_to_add]
-        cur_month.add(1, 'month')
-        date = cur_month.clone().startOf('week')
+      fillerMonths = $scope.getMonths(num_empty_months_to_add, cur_month)
```
lukeferguson: Are the filler months always additional months to be added at the end? What happens if the API returned a data set where there was no availabloyt between range i.e. some aviale in jan, some in march, but none in feb?

TolgaCura: Yes, the filler months always get appended to the end. The code uses the last available date from the 'dates' object passed  into the processDates function to calculate the number of months to display. The filler months will then append on to the end if necessary. In your example, January, February and March should be displayed with February not containing available dates

PR #453, opened 2016-09-15
Mark aria-hidden as true when the widget is covered by a modal

edclements:


edclements:
It would be better if bootstrap ui handled this

lukeferguson:
@edclements just for for reference, the `SettingsService` has a `isModalOpen` open that uses the `$uibModalStack` to determine if a modal is open rather than doing any DOM checking yourself

edclements:
@lukeferguson that sounds potentially useful. Does it account for whether or not the directive is inside the modal though?

PR #474, opened 2016-09-27
remove not needed logic for cancel booking which acted buggy, remove …

MarkShakespeare:
…'#' from href to prevent page reload


src/member/javascripts/controllers/member_bookings.js.coffee
```diff
@@ -64,8 +64,7 @@ angular.module('BBMember').controller 'MemberBookings', ($scope, $uibModal,
 
   $scope.cancelBooking = (booking) ->
     index = _.indexOf($scope.upcoming_bookings, booking)
```
lukeferguson: you can remove this line too @MarkShakespeare

MarkShakespeare: That's needed to push the booking back into the array at the same position if the deletion fails, or does that not matter?

PR #496, opened 2016-10-05
[CSP-4] Option added to not display a default infowindow now - just t…

gwenau:
Two main improvements:

1) Option added to not display a default info window on the map so that only the markers are displayed 

2) When a marker on the map is selected, the corresponding panel will open up as well (previously, the panel could be displaying another company and this can cause confusion)


src/core/javascripts/controllers/map.js.coffee
```diff
@@ -469,6 +469,8 @@ angular.module('BB.Controllers').controller 'MapCtrl', ($scope, $element,
     google.maps.event.trigger($scope.myMap, 'resize')
     $scope.myMap.fitBounds(localBounds)
 
+    return if $scope.options and $scope.options.no_default_location_details is true
```
lukeferguson: Hey @gwenau - I would refactor this code into a method, something called `openDefaultMarket` or something like that, moving the existing logic and your return statement into that.  This will help better encapsulate this behaviour.

gwenau: Hi @lukeferguson 

Thanks for the feedback :)

Please review the updated code and let me know if there is something that I've missed.

PR #498, opened 2016-10-06
[WAIT-93] Modified basket Controller

Joensg:
Modified the controller for 'applyCoupon' and 'applyDeal' to first update the basket before applying the coupon/deal.

**Change Notes:** Modified the controller for 'applyCoupon' and 'applyDeal' to first make a call to update the basket on the server before applying the coupon/deal code.
**Note:** Please see https://bookingbug.atlassian.net/browse/WAIT-93
This fix has also been proposed locally for 'waitrose' project at https://github.com/BookingBug/bookingbug-angular-bespoke/pull/289
The issue can be observed at the following url : https://bespoke.bookingbug.com/waitrose/staging/finchley.html

<h3 align="center">
Before:  implementing changes :
</h3>

**Screencast:**

**Notes:** It was observed on the 'waitrose' project that after applying a coupon/deal code, the individual ticket details that were previously entered(especially noticeable for booking more than a single ticket at a time) like first_name, last_name, email were being cleared.

This was due to the fact that applying a coupon/deal code would essentially just send the basket coupon/deal code over to the server, without first updating the basket on the server. So, the server was returning with a basket which didn't have those details. Thus, the front-end journey would discard its existing basket and use the one from the server with the coupon applied.

<h3 align="center">
After:  implementing changes :
</h3>

**Screencast:**

**Notes:** The basket controller was modified to first make a call to update the basket on the server before making a call to apply the coupon/deal code. This now reflects correctly in the front-end journey.


src/core/javascripts/controllers/basket.js.coffee
```diff
@@ -261,17 +273,23 @@ angular.module('BB.Controllers').controller 'BasketList', ($scope, $rootScope,
...
-        AlertService.add("danger", { msg: err.data.error })
+
+    # first update the basket item if it exists
+    promises = []
+    if params.bb.basket.items
+      promises.push BBModel.Basket.$updateBasket($scope.bb.company, params.bb.basket)
+    $q.all(promises).then () ->
+      BBModel.Basket.$removeDeal($scope.bb.company, params).then (basket) ->
+
+        for item in basket.items
```
lukeferguson: @Joensg this block is repeated 3 times which defies DRY coding convention - if you can refactor to a local method that would be good

src/core/javascripts/controllers/basket.js.coffee
```diff
@@ -210,17 +210,23 @@ angular.module('BB.Controllers').controller 'BasketList', ($scope, $rootScope,
...
-      if err and err.data and err.data.error
-        AlertService.clear()
-        AlertService.add("danger", { msg: err.data.error })
-      loader.setLoaded()
+
+    # first update the basket item if it exists
+    promises = []
+    if params.bb.basket.items
+      promises.push BBModel.Basket.$updateBasket($scope.bb.company, params.bb.basket)
+    $q.all(promises).then () ->
```
lukeferguson: $q.all should only be used when there's more than one promise to resolve which doesn't appear to be case here.

src/core/javascripts/controllers/basket.js.coffee
```diff
@@ -210,17 +210,23 @@ angular.module('BB.Controllers').controller 'BasketList', ($scope, $rootScope,
...
+      BBModel.Basket.$applyCoupon($scope.bb.company, params).then (basket) ->
+        for item in basket.items
+          item.storeDefaults($scope.bb.item_defaults)
+        basket.setSettings($scope.bb.basket.settings)
+        $scope.setBasket(basket)
+        loader.setLoaded()
+      , (err) ->
+        if err and err.data and err.data.error
+          AlertService.clear()
+          AlertService.add("danger", { msg: err.data.error })
```
lukeferguson: We should avoid directly showing error messages from the API to the user.  I'd create a set of errors in the ErrorService for coupons/deals.

Joensg:
@lukeferguson I have considered your requested changes and updated accordingly. Could you please review the modifications? Thanks! :)

Joensg:
@luke Should the `$scope.items = $scope.bb.basket.items` be added into the new method and be done for all three scenarios, that is applyCoupon, applyDeal and removeDeal?

Also, should `loader` function be called in all the cases as it is being called in the case of applyCoupon?

PR #516, opened 2016-10-17
use member model cancel method instead so member is refreshed…

MarkShakespeare:
…, raise alert after cancellation


MarkShakespeare:
Actually would it be better to do it like this using the model method instead ? - 

member_booking_cancel

PR #519, opened 2016-10-19
admin move_booking modal close button for simpler ux, fix confirmation padding 

MarkShakespeare:


src/admin-booking/templates/admin_move_booking_popup.html
```diff
@@ -4,5 +4,6 @@
 <div class="modal-body">
   <div bb-admin-move-booking="{{config}}" class="container-fluid"></div>
 </div>
-
-
+<div ng-if="bb.moved_booking" class="modal-footer">
+  <button type="button" class="btn btn-primary pull-right" ng-click="cancel()">Close</button>
```
lukeferguson: If we add a close button here, should we add one to all modals? They should behave consistently

MarkShakespeare: updated to other modals as well

PR #532, opened 2016-11-14
scope.login defined twice within same file

Joensg:
**Change Note:**
Renamed variable name which was also defined with the same name as a function later within the same file.

'$scope.login' had been defined as a variable and also as a function later within the same file.
Hence, the function definition was essentially overriding the previously defined variable.

Note that this was observed on both 'admin' and 'admin-dashboard'.

### DIFF
--- a/src/admin-dashboard/javascripts/login/directives/admin_dashboard_login.js.coffee
+++ b/src/admin-dashboard/javascripts/login/directives/admin_dashboard_login.js.coffee
@@ -117,7 +117,7 @@ angular.module('BBAdminDashboard.login.directives').directive 'adminDashboardLog
         $scope.template_vars.show_login = false
         companySelection($scope.user)
 
-      $scope.login = (isValid) ->
+      $scope.login_function = (isValid) ->
         if isValid
           $scope.template_vars.show_loading = true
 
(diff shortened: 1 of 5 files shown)


src/admin-dashboard/javascripts/login/directives/admin_dashboard_login.js.coffee
```diff
@@ -117,7 +117,7 @@ angular.module('BBAdminDashboard.login.directives').directive 'adminDashboardLog
         $scope.template_vars.show_login = false
         companySelection($scope.user)
 
-      $scope.login = (isValid) ->
+      $scope.login_function = (isValid) ->
```
lukeferguson: method names should describe what they do, not what they are! i'd rename the login var instead

lukeferguson:
@Joensg - something I've just realised - login_form shared it's name with the login_form created by ng-form - we'll have to rename it again to avoid sharing scope - I think `login_data` would be appropriate! Sorry for missing this!

Joensg:
@lukeferguson ok, thanks!
I created a PR for it at https://github.com/BookingBug/bookingbug-angular/pull/578 .

PR #535, opened 2016-11-15
calendar accessibility 

MarkShakespeare:
added ability to tab through calendar accordion, skips unavailable slots and focus on continue button after selecting a slot

### DIFF
--- a/src/core/javascripts/controllers/time_range_list.js.coffee
+++ b/src/core/javascripts/controllers/time_range_list.js.coffee
@@ -27,7 +27,7 @@
 ####
 
 
-angular.module('BB.Directives').directive 'bbTimeRanges', ($q, $templateCache, $compile, $timeout) ->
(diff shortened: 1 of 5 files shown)


src/core/javascripts/controllers/time_range_list.js.coffee
```diff
@@ -35,6 +35,10 @@ angular.module('BB.Directives').directive 'bbTimeRanges', ($q, $templateCache, $
   transclude: true
   controller : 'TimeRangeList'
   link: (scope, element, attrs, controller, transclude) ->
+    scope.$watch 'selected_slot', ->
+      if scope.bb.current_item.time
+        btn = element[0].parentNode.getElementsByClassName('btn-primary')
```
lukeferguson: I'd give the 'continue' button an id and use that to get the element - otherwise adding another btn-primary will break this.  Rather than watching the selected_slot too - can this not be bound to the highlight slot method? We should generally avoid introducing watches given their performance burden.

PR #536, opened 2016-11-17
[IMPL-1399] renamed '$scope.filters.hide_sold_out_events' to '$scope.filters.hide_fully_booked_events'

Joensg:
**Change Notes:**
- Renamed '$scope.filters.hide_sold_out_events' to '$scope.filters.hide_fully_booked_events' in the javascript file and made corresponding changes in public-booking/templates file.
- $scope.filters.hide_fully_booked_events initially set to 'false'.

### DIFF
--- a/src/core/javascripts/controllers/event_list.js.coffee
+++ b/src/core/javascripts/controllers/event_list.js.coffee
@@ -56,7 +56,7 @@ angular.module('BB.Controllers').controller 'EventList', ($scope, $rootScope, Ev
   $scope.pick = {}
   $scope.start_date = moment()
   $scope.end_date = moment().add(1, 'year')
-  $scope.filters = {hide_sold_out_events: true}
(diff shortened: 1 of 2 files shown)


src/core/javascripts/controllers/event_list.js.coffee
```diff
@@ -20,6 +20,7 @@
 * @param {hash}  bbEvents A hash of options
 * @property {integer} total_entries The event total entries
 * @property {array} events The events array
+* @property {boolean} hide_fully_booked_events - Events that are fully booked but have a waitlist will be hidden. Events that are fully booked and DO NOT have a waitlist are ALWAYS hidden.
```
lukeferguson: The property description should detail its effect and as a boolean, it's default value, i.e. 'Hide fully booked events (i.e. events that either only waitlist spaces left or no spaces at all). Default is false.'

PR #544, opened 2016-11-21
collaspe => collapse

MarkShakespeare:


lukeferguson:
Marking as temp-closed - will roll into a future major release @MarkShakespeare

PR #570, opened 2016-12-12
IMPL-1388 admin booking & admin block: possibility to change person a…

damianpudlo:
…sset across various callendar views


src/core/javascripts/controllers/person_list.js.coffee
```diff
@@ -114,8 +114,7 @@ angular.module('BB.Controllers').controller 'PersonList', ($scope, $rootScope,
         # OR if the person has been passed into item_defaults, skip to next step
         if (items.length is 1 and $scope.bb.company.settings and $scope.bb.company.settings.merge_people)
           person = items[0]
-        if $scope.bb.item_defaults.person
-          person = $scope.bb.item_defaults.person
```
damianpudlo: It was added in last month but it makes it impossible to change currently selected person asset. 
I removed that - is that ok?

PR #579, opened 2016-12-30
bb-scroll-to : scroll to top of the calendar.

Joensg:
**Change Notes:**
- Added a local helper function `isElementInView()` that would return true if the element was visible within the viewport.

**Description:**
Issue noticed on MHM, please see https://bookingbug.atlassian.net/browse/IMPL-351 .

Given that : 
- smaller screen sizes (tablet/mobile)
- on the calendar page
- the webpage has a large header/footer

It was noticed that on selecting a time slot via the drop down accordion, the page remains at the current position(header/footer) on accordion-collapse.
I believe that it is not expected behaviour and that instead, on selecting the time slot, we would expect the selected time slot to have focus and be brought into view.

Also, @lukeferguson please consider the following : 
the variable `always_scroll` is being set using the `bb-always-scroll` attribute at https://github.com/BookingBug/bookingbug-angular/pull/579/files#diff-c4ff8970dc82e52659753632608331f6R70
and is being used here https://github.com/BookingBug/bookingbug-angular/pull/579/files#diff-c4ff8970dc82e52659753632608331f6R93
Is the `always_scroll` variable redundant? Shouldn't we always be scrolling to the element by default as long as a `scroll_to_element` has been set?


src/core/javascripts/directives/main.js.coffee
```diff
@@ -81,6 +81,8 @@ angular.module('BB.Directives').directive 'bbScrollTo', ($rootScope, AppConfig,
     scrollToCallback = (evnt) ->
       if evnt == "page:loaded" && scope.display && scope.display.xs && $bbug('[data-scroll-id="'+AppConfig.uid+'"]').length
         scroll_to_element = $bbug('[data-scroll-id="'+AppConfig.uid+'"]')
+      else if evnt == "time:selected"
```
lukeferguson: @Joensg: Rather than reacting to an event and targeting specific element, the directive should be invoked on the element, i.e. `<div bb-time-ranges bb-scroll-to="time:selected"></div>`

PR #595, opened 2017-01-13
Core 1175

diegomary:
Please see also the following pull request for Generator BB:

https://github.com/BookingBug/generator-bookingbug/pull/32


build/booking-widget/bower.json
```diff
@@ -1,41 +0,0 @@
-{
```
damianpudlo: the following files removal should be confirmed with Ed (I believe he needs them)

build/booking-widget/bower.json
build/booking-widget/src/stylesheets/_bb-variables.scss
build/booking-widget/src/stylesheets/_bootstrap-variables.scss
build/booking-widget/src/stylesheets/_theme.scss
build/booking-widget/src/stylesheets/main.scss

src/core/javascripts/constants/constant.js.coffee
```diff
@@ -0,0 +1,18 @@
+angular.module('BB').constant 'ROUTE_STATES',
```
diegomary: Hi Damien, I agree and we can change it, however Luke decided that for constants we could  use all capital name C style.

PR #598, opened 2017-01-13
[X] - fix print preview for google maps image

TolgaCura:


TolgaCura:
Before Fix
screen shot 2017-01-13 at 15 20 41

After Fix
screen shot 2017-01-13 at 15 22 49

PR #612, opened 2017-01-19
[X] - place viewport elements under bb id, corrected invalid styling

TolgaCura:


TolgaCura:
Before 
screen shot 2017-01-19 at 15 09 51

After 
screen shot 2017-01-19 at 15 11 21

PR #623, opened 2017-01-26
tidy up admin_move_booking_popup, 

C-E-Rios:
prevent revertFunc being called when event moved via drag/drop


src/admin-booking/javascripts/directives/admin_move_booking_popup.js.coffee
```diff
@@ -1,32 +1,37 @@
...
+    $scope.config.company_id ||= $scope.company.id if $scope.company
+    
+    $scope.config.item_defaults = angular.extend
+      merge_resources: AdminBookingOptions.merge_resources
+      merge_people: AdminBookingOptions.merge_people
+    , config.item_defaults
+    
+    $scope.cancel = () ->
+      event.preventDefault()
+      event.stopPropagation()
```
damianpudlo: could those events be removed?

PR #636, opened 2017-02-07
[Core-1379] Studio UI - Improve UX for datepicker and timepicker in Edit Block/Create Block Modal.

Joensg:
**Change Notes:**
- the time input fields can now _only_ be changed by using the top and bottom chevrons.
- the date input field can now _only_ be changed by first clicking on the input or calendar icon and selecting a date from the datepicker popup.
- the week number is now hidden on the uib-datepicker.
- the language picker is now hidden on quick_pick modal when on admin interface.
- implemented style changes as has been proposed in the PR review.
- Can demo and test this implemented changes on studio dev at : http://bespoke.bookingbug.com/studio/development/ (see ticket below for test login credentials)
- Please see https://bookingbug.atlassian.net/browse/CORE-1379 for more details.

## Screenshot before fix:
_(Note how green correct tick mark is out of place, the date and time inputs are editable as text)_

_(Note how two green top chevrons are being hidden on smaller screen sizes)_

_(Note how the date and time inputs are editable as text)_

_(Note how four green top chevrons are being hidden on smaller screen sizes)_

_(Note how week number is being shown on the uib-datepicker)_

## Screenshot with changes implemented:
_(Note how green correct tick mark is now in place, the date and time inputs are not editable as text)_

_(Note how two green top chevrons are now not being hidden on smaller screen sizes)_

_(Note how the date and time inputs are now not editable as text, the language picker is hidden)_

_(Note how four green top chevrons are now not being hidden on smaller screen sizes, the language picker is hidden)_

_(Note how week number is now hidden on the uib-datepicker)_

_(Note how the invalid inputs are now highlighted with label and border of red colour)_


src/admin-booking/stylesheets/date_picker.scss
```diff
@@ -1,8 +1,20 @@
...
 
+}
 
-}
\ No newline at end of file
+.form-group {
+  [bb-date-time-picker]{
+    .bb-date-picker {
+      input {
+        &.form-control[readonly] {
```
lukeferguson: Does this rule need to limited to `has-success` selector too?

src/admin-dashboard/stylesheets/main.scss
```diff
@@ -38,3 +38,27 @@ html,body {
...
+    &.schema-form-duration {
+      position: relative;
+    }
+    &.schema-form-duration > .form-control-feedback {
+      top: 23px;
+    }
+  }
+
+  .input-group {
+    input[uib-datepicker-popup] ~ span > button {
```
lukeferguson: Similarly, should it only be green when it's valid?

src/core/templates/bb_date_time_picker.html
```diff
@@ -1,21 +1,22 @@
...
       <span class="input-group-btn" ng-click="$event.preventDefault(); $event.stopPropagation(); opened=!opened;">
         <button class="btn btn-default" title="Select date"><span class="fa fa-calendar"></span></button>
       </span>
     </div>
   </div>
-  <div class="col-sm-5" ng-show="!dateOnly">
+  <div class="col-xs-5" ng-show="!dateOnly">
     <div uib-timepicker ng-model="datetimeWithNoTz" show-meridian="showMeridian"
-      minute-step="minuteStep" max="maxDateClean" min="minDateClean" readonly-input="bbDisabled"></div>
+      minute-step="minuteStep" max="maxDateClean" min="minDateClean" readonly-input="true"></div>
```
lukeferguson: Should we still let this be configurable (e.g. if the time picker is used elsewhere?)

src/admin-booking/stylesheets/date_picker.scss
Joensg: the is already been taken care of in the AdminLTE-angular-sass at : https://github.com/BookingBug/AdminLTE-angular-sass/blob/master/src/stylesheets/_forms.scss#L32

src/core/templates/bb_date_time_picker.html
Joensg: @lukeferguson the `uib-timepicker` is being used in conjunction with the bb-date-picker.
I believe that the uib-timepicker is not being set to be shown except for the admin interfaces.
If setting the `readonly` to be configurable, it results in not a very good ux, since in the case where the time-step has been set to 5 mins and the user enters 8 mins and clicks save, it would still save it as 5 mins or 10 mins.
If the timepicker needs to be used elsewhere, it could be used by simply changing the uib-timepicker options that are being passed in.

src/admin-dashboard/stylesheets/main.scss
Joensg: @lukeferguson if I add the `.has-success` selector to it, it fails to do it for the create Block modal, since it lacks any such class. Please see https://github.com/BookingBug/bookingbug-angular/blob/master/src/admin-booking/templates/quick_pick.html

src/admin-booking/stylesheets/date_picker.scss
Joensg: @lukeferguson if I add the `.has-success` selector to it, it fails to do it for the create Block modal, since it lacks any such class. Please see https://github.com/BookingBug/bookingbug-angular/blob/master/src/admin-booking/templates/quick_pick.html

lukeferguson: Do any of the ui-boostrap components automatically apply it? 

Just need to ensure we're consistent with our application of it - ideally the border colour should be just the bootstrap default, not green

src/core/templates/bb_date_time_picker.html
lukeferguson: Yep, point taken about time steps - just thinking about usage where time steps aren't applicable. Happy to take approach suggested.

src/admin-booking/stylesheets/date_picker.scss
Joensg: @lukeferguson for the Edit modal block at https://github.com/BookingBug/bookingbug-angular/blob/master/src/admin-dashboard/templates/edit_block_modal_form.html , the class `has-success` is being applied automatically from the schema form. While for the Create block form at https://github.com/BookingBug/bookingbug-angular/blob/master/src/admin-booking/templates/quick_pick.html , no such class is being applied. I have now added an `ng-class` to this form and it fixes it.

Regarding the proposed styles, please note that we are in fact setting theses inputs to be `readonly` but these can still be modified using the datepicker-popup or the up/down chevrons and hence we would like to have these inputs to appear as modifiable to the user. The above styles are being added in for that specific purpose, such that only when its the datepicker and the input field is `readonly`, we set it to appear as modifiable.

Also note that the proposed styles above are already been applied/overridden by the following : https://github.com/BookingBug/AdminLTE-angular-sass/blob/master/src/stylesheets/_forms.scss#L27

Joensg: @lukeferguson I have now implemented the proposed style changes and updated the screenshots accordingly. You could also test the changes on studio dev.

PR #639, opened 2017-02-09
[CORE-1454] bbForm fix

### LINKED ISSUE #638: Multiple interfaces for validating forms is confusing
Some components leverage the `ValidatorService` by adding it `$scope`, whereas others rely on the `submitForm()` method introduced by `bbForm`.

A single interface for form validation would facilitate consistent implementation.

lukeferguson:
Fixed issue where form validation was not triggered when attempting to proceed to next step.

Fixes #638


src/core/javascripts/directives/steps/bb_deals/bb_deals.controller.js.coffee
```diff
@@ -5,6 +5,9 @@ angular.module('BB.Controllers').controller 'DealList', ($scope, $rootScope, $ui
   FormDataStoreService.init 'DealList', $scope, [ 'deals' ]
   loader = LoadingService.$loader($scope).notLoaded()
 
+  console.warn('Deprecation warning: validator.validateForm() will be removed from bbDealList in an upcoming major release, please update your template to use bbForm and submitForm() instead. See https://github.com/bookingbug/bookingbug-angular/issues/638')
```
lukeferguson: I've added it here as I found bespoke projects that still expect `validator` to be in this directives scope.

src/core/javascripts/directives/inputs/bb_form.directive.spec.js.coffee
```diff
@@ -190,17 +189,17 @@ describe 'bbForm directive', ->
 
       return
 
-    it 'checkReady will be not triggered ', ->
+    it 'checkReady will be triggered ', ->
```
lukeferguson: I've changed the spec to reflect changes that will default route to be used when route isn;t explicitly set.

PR #652, opened 2017-02-20
[IMPL-1870] [master] disable add button('right-chevron') when time range greater than  max days in advance.

Joensg:
**Change Notes:**
- [IMPL-1870] ability to disable the add button('right-chevron') on week calendar when time range is more than the max days in advance.
- display popover when trying to click on add button('right-chevron') when max date in advance reached.
- datepicker-popup on week calendar page made to respect min, max date.
- show the 'jump to next available date' button on week calendar page only if not more than max date in advance.
- changed popover text as requested by peer review.

### Screenshot of changed popover text
_(Note how the popover text value now uses durations)_

### Case 1: Time range length = 7, Max advance date = March 6th 2017
_(Note how the right-chevron is correctly disabled)_

### Case 2: Time range length = 5, Max advance date = March 6th 2017
_(Note how the right-chevron is correctly disabled)_

### Case 3: Time range length = 3, Max advance date = March 6th 2017
_(Note how the right-chevron is correctly disabled)_

### Case 4: Time range length = 1, Max advance date = March 6th 2017
_(Note how the right-chevron is correctly disabled)_


src/core/javascripts/directives/steps/bb_time_ranges/bb_time_ranges.controller.js
```diff
@@ -157,6 +160,38 @@ angular.module('BB.Controllers').controller('TimeRangeList', function ($scope, $
...
+    /***
+     * @ngdoc method
+     * @name setMinMaxDate
+     * @methodOf BB.Directives:bbTimeRanges
+     * @description
+     * Set min, max date and time range based on min/max advance datetime of the selected service
+     *
+     * @param {object} current_item The current item
+     */
+    var setMinMaxDate = function (current_item) {
```
lukeferguson: I wouldn't pass in the current item - when would it be something that isn;t?

src/core/javascripts/directives/steps/bb_time_range_stacked/bb_time_range_stacked.controller.js
```diff
@@ -95,9 +99,42 @@ angular.module('BB.Controllers').controller('TimeRangeListStackedController', fu
...
+
+        // has a service been selected?
+        if (current_item.service && !$scope.options.ignore_min_advance_datetime) {
+
+            $scope.min_date = current_item.service.min_advance_datetime;
+            $scope.max_date = current_item.service.max_advance_datetime;
+
+            // date helpers for use by datepicker-popup
+            $scope.min_date_js = $scope.min_date.toDate()
+            $scope.max_date_js = $scope.max_date.toDate()
```
lukeferguson: use camel case for the new var and covert existing one if it was unused in the default template

src/core/javascripts/directives/steps/bb_time_range_stacked/bb_time_range_stacked.controller.js
```diff
@@ -39,6 +39,10 @@ angular.module('BB.Controllers').controller('TimeRangeListStackedController', fu
             }
         }
 
+        $scope.options.ignore_min_advance_datetime = $scope.options.ignore_min_advance_datetime ? true : false;
+
+        setMinMaxDate($scope.bb.current_item);
```
lukeferguson: The service might have changed from initial load (i.e. use case where service selection is on the same step of the calendar) so the min/max dates and validity flags need to be set in the `loadData` method.  This can be done before the time data comes back.

PR #662, opened 2017-02-28
[CORE-1511] - enable mobile date picker to update time range calendar

TolgaCura:


TolgaCura:
When creating a new booking journey using SDK, the calendar date picker (which is only displayed on mobile) does not update the time range accordion. This PR should correct this bug.

PR #663, opened 2017-03-03
Way to detect if the journey is a parent service journey (service first)

DeanPalmer:
We have decided that this is the best way to decide whether to run the code that sets the selected_service as an item default. It simple checks the order in which the service and company templates are in the AllSteps array. Its crude but will work if people do not stray from our naming conventions for template names and if not it will run the code as original so it will not cause any issues.


src/core/javascripts/directives/bb_widget/bb_widget_init.service.js
```diff
@@ -148,9 +148,34 @@
...
+                    for (let stepName of stepNames) {
+                        if (step.page.includes(stepName)) return true;
+                    }
+                });
+
+                if (stepMatched !== -1) return stepMatched;
+
+                return null
+            }
+            //Checks the order in which a service page sits in bb.allSteps in comparson with the company page
```
WesleyPumpkinhead: typo for comparison

PR #679, opened 2017-03-28
[IMPL-2180] emit event on dayList loaded.

Joensg:
**Change Note:**
- [IMPL-2180] emit event on bbDayList loaded so that it could be used by `bb-scroll-to`.


src/core/javascripts/directives/steps/bb_day_list/bb_day_list.controller.js
```diff
@@ -87,6 +87,8 @@ angular.module('BB.Controllers').controller('DayList', function ($scope, $rootSc
                 $scope.weeks = _.groupBy($scope.day_data, day => day.date.week());
                 $scope.weeks = _.toArray($scope.weeks);
 
+                $scope.$emit('bbDayList:loaded');
```
lukeferguson: Only components that sit higher in the scope hierarchy - is this intended behaviour? `$rootScope.emit` might be more appropriate.

PR #680, opened 2017-03-28
[IMPL-2004] added an option 'no_route' to Purchase

Joensg:
**Change Note:**
- [IMPL-2004] added an option 'no_route' which enables not calling decideNextPage on Purchase page.


lukeferguson:
@Joensg can you please liaise with @MarkShakespeare - might be some duplication of effort here (See CORE-1085)

Joensg:
Discussed with Luke and we agreed that IMPL-2004 would need to wait until CORE-1085 is complete. Thus closing this PR.

PR #682, opened 2017-03-29
Fix selected timeslot on Basket object getting set to current time af…

jonw-bb:
…ter updating basket (e.g. when progressing from Calendar page to next step)


jonw-bb:
Investigating with Damian a better way to fix issue...

PR #684, opened 2017-03-30
[IMPL-2004] fix 'bb-scroll-to' to scroll correctly on a modal.

Joensg:
**Change Note:**
- [IMPL-2004] fix 'bb-scroll-to' to scroll correctly on a modal.

**Details:**
It was noticed that if `bb-scroll-to` was set on an element located on a page loaded inside a modal, rather than scrolling to the appropriate element, the page was scrolling in the greyed background of the modal. This was due to the code calling the scrollTop function on the wrong context(the html, body) instead of on the modal(when the modal is open).


DeanPalmer:
Looks fine but can you please use the scrollIntercepter service to perform the scroll?

PR #686, opened 2017-03-31
Core 1562- Studio booking summary view accessibility improvement

terencemutizwa:
added tab indexes and aria-labels in basket summary view


terencemutizwa:
accessible studio booking confirmation 
booking_confirmation

jthomas1:
Since you're doing accessibility improvements I would recommend changing the colours of the labels on the confirmation page from the very light grey to something darker because they are very hard to read at that contrast ratio.

terencemutizwa:
ok noted, will make the changes, i thought they were disabled handles initially

PR #697, opened 2017-04-10
Core 1619 - accessible booking calendar

terencemutizwa:


terencemutizwa:
accessible_calendar
working accessible calendar, tab stop into and out of calendar grid, navigate between cells and within accordion options using arrow keys, space/enter to select, esc key to escape

PR #722, opened 2017-04-18
[IMPL-2330] Add service to allow disabling of Serve Next and other bu…

WesleyPumpkinhead:
…ttons while waiting for response from API


src/queue/javascripts/services/admin_queue_serve_now.js
```diff
@@ -0,0 +1,11 @@
+angular.module('BBQueue.services').factory('AdminQueueLoading', () => {
```
damianpudlo: it's an object instance so should be lowercase 'adminQueueLoading'

src/queue/javascripts/services/admin_queue_serve_now.js
```diff
@@ -0,0 +1,11 @@
...
+    let loadingServerInProgress = false;
+    return {
+        getLoadingServerInProgress: function () {
```
damianpudlo: for functions returning boolean it's better to use 'is' prefix instead of 'get'
could we change that to 'isLoading'

src/queue/javascripts/controllers/queue_server.js
```diff
@@ -1,8 +1,11 @@
 let QueueServerController = ($scope, $log, AdminQueueService, ModalForm, BBModel, CheckSchema,
-                             $uibModal, AdminPersonService, $q, AdminQueuerService, Dialog, $translate) => {
+                             $uibModal, AdminPersonService, $q, AdminQueuerService, AdminQueueLoading, Dialog, $translate) => {
+
+    $scope.AdminQueueLoading = AdminQueueLoading;
```
damianpudlo: please expose only methods to the view instead of whole service

scope.setLoadingServerInProgress = adminQueueLoading.setLoadingServerInProgress;
scope.isLoadingServerInProgress = adminQueueLoading.isLoadingServerInProgress;

src/queue/javascripts/controllers/queue_server.js
```diff
@@ -1,8 +1,11 @@
...
+
+    $scope.loadingServer = false;
```
damianpudlo: Existence of this variable seems to be duplication since with have adminQueueLoading service.
Could we remove it?

WesleyPumpkinhead: no, this var is used for something else

PR #746, opened 2017-05-08
Core 1808

flaviotulino:
There was a little problem after editing a client, because after hitting the save button (which goes to the summary step) and then the back button, were showed the old data and not the updated ones.
This issue has been solved forcing typehead to re-perform the request even if the search term has not changed: in this way the search is always triggered and the data are always updated.

I'm considering to find another option instead of clearing the cache via the halClient, but, as is stands, now it works.


flaviotulino:
Added a minor fix to avoid false positive errors in the customers search bar

src/admin-booking/javascripts/controllers/clients.js
```diff
@@ -36,19 +36,72 @@
...
 
+
+        $scope.client.isSelected = false;
+
         $scope.selectClient = (client, route) => {
             $scope.setClient(client);
             $scope.client.setValid(true);
-            return $scope.decideNextPage(route);
+
+            DataBucket.put('client', client);
```
lukeferguson: @flaviotulino is the DataBucket necessary when value is held in parent scope by bbWidget? Also consider the `FormDataStoreService` which is used already to persist values on scope.

PR #752, opened 2017-05-11
keyboard navigable language picker 

terencemutizwa:
key_nav


C-E-Rios:
Language picker should behave similarly to 'userAccount' dropdown. ie: Background should darken.

terencemutizwa:
lang

PR #757, opened 2017-05-15
[CORE-1968] fixed doc generation errors

flaviotulino:
All the `/*` at the beginning of ngdoc blocks have been refactored to `/**`.
All the `<example>` tags were missing the `module` attribute referencing the modules itselves

**This issue is a sub-task of the **[CORE-1139](https://bookingbug.atlassian.net/browse/CORE-1139)** task**


src/core/javascripts/i18n/directives/bb_time_zone_select.component.js
```diff
@@ -3,20 +3,18 @@
     'use strict';
 
     /**
-     * @ngdoc component
+     * @ngdoc directive
```
jthomas1: I guess gulp-ngdocs doesn't know about the component syntax?

flaviotulino: Unfortunately no.

jthomas1: Ok, I am going to create a ticket so we can update it to add this functionality.

PR #759, opened 2017-05-15
[Core 1940]

flaviotulino:
Added the layout attribute to bb-map directive to manage the default/radio behaviour.
If the radio value is set, when the user selects a store his choice is highlighted via a checked radio, ; the "select" button is hidden because the store is selected by default and a "continue" button is displayed below the map.

**UPDATE** 
There is a little bit of confusion regarding the relationship between two issues: CORE-1940 and CORE-1948.
They are strictly related but one of them is blocked by the Opening Hours feature we need to implement, while the other one is just an enhancement for the selection.


src/public-booking/templates/map.html
```diff
@@ -193,7 +197,7 @@
               class="store-details"
               ng-click="openMarkerInfo(marker)">
               <li ng-bind-html="marker.company.address | address_multi_line"></li>
-              <li>
+              <li ng-if="marker.company.address.phone">
```
lukeferguson: @flaviotulino we generally favour `ng-show` given that it doesn't create new scope.

src/public-booking/templates/map.html
```diff
@@ -232,8 +237,13 @@
...
-    class="bb-step-navigation">
-  </div>
-</div>
+
+  <button
+    ng-if="layout.radio && currentMarker"
+    type="button"
+    ng-click="selectItem()"
+    class="btn btn-primary pull-right"
+    translate="PUBLIC_BOOKING.CALENDAR.NEXT_BTN">
```
lukeferguson: Please add a new unique key. Check out the wiki for future reference too ;)

https://github.com/BookingBug/bookingbug-angular/wiki/1.1-i18n

lukeferguson:
Please mark PR's with the 'ready to review' label too :)

src/public-booking/templates/map.html
flaviotulino: Agree. Just changed it

flaviotulino:
Map marker have been labelled with the number of the store, to make easier to understand which one has been selected.
screen shot 2017-05-17 at 17 16 54
screen shot 2017-05-17 at 17 17 06


Passing layout='radio' to the bb-map directive, the result is the following
screen shot 2017-05-17 at 17 27 19

flaviotulino:
@lukeferguson could you please take a look to this update?

PR #763, opened 2017-05-19
[CORE-1968] Fix documentation deploy process

flaviotulino:
A new gulp task has been created to handle the documentation generation and its deploy onto an s3 bucket.
All the files are uploaded under https://angular.bookingbug.com/documentation/<VERSION>, which contains all the html partials and the setup js (generated by the ngdocs task)


gulp-tasks/tasks-config/docs.js
```diff
@@ -32,10 +32,79 @@
                 scripts: ['test/examples-archive/booking-widget.js']
             };
             return gulp.src('src/*/javascripts/**')
-                .pipe(gulpIf(/.*coffee$/, gulpCoffee().on('error', gulpUtil.log)))
-                .pipe(gulpDocs.process(options))
-                .pipe(gulp.dest('./docs'));
+            .pipe(gulpIf(/.*coffee$/, gulpCoffee().on('error', gulpUtil.log)))
```
jthomas1: Presumably we don't need this line about coffeescript anymore since the sdk is es6 only now?

gulp-tasks/tasks-config/docs.js
```diff
@@ -32,10 +32,79 @@
...
+            .pipe(gulpDocs.process(options))
+            .pipe(gulp.dest('./docs'));
         });
+
+        gulp.task('docs:sdk-upload', function () {
+            const fs = require('fs');
```
jthomas1: put the imports at the top of the file and while you're there you can change the existing ones from `var` to `const`

PR #768, opened 2017-05-23
Check if a booking can be cancelled and disable the cancel button if not.

jthomas1:
I have modified the purchase.html template of public booking so that if the user is viewing a booking that falls within the minimum cancellation period the cancel button will be disabled and a message displayed explaining why.

This enhancement came off the back of fixing the issue for the Capital One journey: https://bookingbug.atlassian.net/browse/IMPL-2411

If this needs changing anywhere else just let me know and I will update.


src/public-booking/templates/purchase.html
```diff
@@ -134,13 +134,18 @@
 
                 <div class="row">
 
-                  <div class="col-sm-4">
+                  <div class="col-sm-4 form-group">
```
lukeferguson: Keep form-groups separate from col definition @jthomas1

jthomas1: So wrap the button and span in another element inside the column?

```
<div class="col">
    <span class="form-group">
....
    </span>
</div>
```

PR #769, opened 2017-05-24
[FIX] IMPL-2344 - Disable left/back arrow on Calendar if no previous dates…

jonw-bb:
… are viewable.

The 'disabled' attribute on the back arrow is set by `is_subtract_valid` boolean flag. Currently this is calculated by looking at whether dates are in the past. This does not account for dates which are today or in future and are non-bookable.

The first bookable date is assigned to `min_date` variable and the calendar will not allow you to view dates before this even they are in the future. The back arrow button is marked as active; it simply doesn't do anything when clicked.

Change made so that `is_subtract_valid` is now calculated by looking at `min_date` instead of current day. `min_date` value is assigned with `current_item.service.min_advance_datetime` which in turn is created with `moment().add(this.min_advance_period, 'seconds')` (in /core/javascripts/models/service.js). So `min_date` is not only the first bookable date, it is also never in the past either.


src/core/javascripts/directives/steps/bb_time_ranges/bb_time_ranges.controller.js
```diff
@@ -322,7 +322,8 @@
         var isSubtractValid = function () {
             $scope.is_subtract_valid = true;
 
-            let diff = Math.ceil($scope.selected_day.diff(moment(), 'day', true));
+            let diff = $scope.selected_day.diff($scope.min_date, 'days');
```
TolgaCura: Should include the rounding from previous version

PR #791, opened 2017-06-08
[CORE-2121] Updated broadcastUpdate to accept a state which can be us…

DeanPalmer:
…ed to stop unnecessary loading.


src/core/javascripts/directives/steps/bb_durations/bb_durations.controller.js
```diff
@@ -72,7 +75,7 @@ angular.module('BB.Controllers').controller('DurationList', function ($scope, $a
      */
     $scope.durationChanged = () => {
         $scope.bb.current_item.setDuration($scope.duration.value);
-        return $scope.broadcastItemUpdate();
+        return $scope.broadcastItemUpdate('Duration');
```
lukeferguson: @DeanPalmer consider using `routeStates` - these are set of constants that define the different states.

PR #792, opened 2017-06-08
Core 173

TolgaCura:
This PR refactors the bbTimeRangeStacked directive to use the new TimeSlotPermutation service. These two directives (with bbMultiServiceSelect) will allow developers to create journey where users can select multiple services and checkout

Below are screenshot of this working

NOTE: the API doesn't support pre/post times for services so bookings map overlap if these are set

First two services selected
screen shot 2017-06-09 at 12 47 00

Slots are displayed for services to be carried out back to back (with a maximum gap between slots set to 0)
screen shot 2017-06-09 at 12 47 28

Confirmation of both services booked one after another
screen shot 2017-06-09 at 12 48 08

Confirmation of booking back to back in admin interface. Note: pre and post times overlap, but service time does not
screen shot 2017-06-09 at 12 56 21


src/core/javascripts/services/time_slot_permutation/time_slot_permutation.service.spec.js
```diff
@@ -0,0 +1,917 @@
...
+                                                                                                                                                                
+            const test_results = TimeSlotPermutationService.permute(test_case);
+            expect(test_results.length).toEqual(factorial(test_case.length));
+        });
+
+        it('generates a valid permutation for an array of string and integers', function () {
+            const test_case = ['@', 3, '#', 15, 24, '~', '+-'];
+
+            const test_results = TimeSlotPermutationService.permute(test_case);
+            expect(test_results.length).toEqual(factorial(test_case.length));
```
jthomas1: Replace `factorial` with the expected value again.

src/core/javascripts/services/time_slot_permutation/time_slot_permutation.service.spec.js
```diff
@@ -0,0 +1,917 @@
...
+            TimeSlotPermutationService.addExistingDateTimes(stacked_items, slots);
+            expect(slots[0][datetime_label].length).toEqual(1);
+            expect(slots[0][datetime_label][0].time).toBe(slot_time);
+        });
+
+        it('does not alter slots array if stacked items are missing date time data', function () {
+            const stacked_items = [
+                {
+                    title: 'service 1',
+                    // datetime:  datetime, // This attribute is being tested (leave commented out)
```
jthomas1: Could reword the comment to `the absense of this attribute is being tested` or remove the attributes entirely to make it explicit that they are not there at all and avoid confusion about commented out code.

src/core/javascripts/services/time_slot_permutation/time_slot_permutation.service.spec.js
```diff
@@ -0,0 +1,917 @@
...
+            ];
+
+            TimeSlotPermutationService.addExistingDateTimes(stacked_items, slots);
+            // previously selected slots are assinged to the begining of the slots array
+            expect(slots[0][datetime_label].length).toEqual(2);
+            expect(slots[0][datetime_label][0].time).toEqual(stacked_items[0].time.time);
+            expect(slots[0][datetime_label][0].selected).toEqual(true);
+        });
+
+        it('slots in current array are pushed back in array when slot previously selected', function () {
```
jthomas1: This could be worded slightly better, doesn't quite make sense in the "it"

src/core/javascripts/services/time_slot_permutation/time_slot_permutation.service.spec.js
```diff
@@ -0,0 +1,917 @@
...
+
+            // previously selected slots are assinged to the begining of the slots array
+            // since this is the 2nd previously selected slot, we're checking the 2nd item in the array
+            expect(slots[1][datetime_label][0].time).toEqual(stacked_items[1].time.time);
+            expect(slots[1][datetime_label][0].selected).toEqual(true);
+        });
+    });
+
+    describe('isItemSlotsValid', function () {
+        it('returns false when not items are passed in', function () {
```
jthomas1: `not items`

src/core/javascripts/services/time_slot_permutation/time_slot_permutation.service.spec.js
```diff
@@ -0,0 +1,917 @@
...
+                    test: 'item 2'
+                }
+            ];
+            const item_slots = [];
+
+            const results = TimeSlotPermutationService.isItemSlotsValid(items, item_slots);
+            expect(results).toBe(false);
+        });
+
+        it('returns true when matching item slots are found and orders slots by time', function () {
```
jthomas1: You are not asserting the other of the slots, only that it returns true. Perhaps you want to do that as a seperate test?

src/core/javascripts/directives/steps/bb_time_range_stacked/bb_time_range_stacked.controller.js
```diff
@@ -325,10 +301,10 @@ angular.module('BB.Controllers').controller('TimeRangeListStackedController', fu
 
             // set the date and time on the stacked items
             while (slot) {
-                for (let item of Array.from($scope.bb.stacked_items)) {
-                    if ((item.service.self === slot.service.self) && !item.date && !item.time) {
-                        item.setDate(day);
-                        item.setTime(slot);
+                for (let itemIndex in $scope.bb.stacked_items) {
+                    if (($scope.bb.stacked_items[itemIndex].service.self === slot.service.self) && !$scope.bb.stacked_items[itemIndex].date && !$scope.bb.stacked_items[itemIndex].time) {
```
jthomas1: For readability perhaps create a variable something like: `current_item = $scope.bb.stacked_items[itemIndex]` and then use that throughout the loop

PR #802, opened 2017-06-15
new configuration for booking max advance days for admin journeys

terencemutizwa:
Problem 
- config settings for max future date allowed for making bookings is set for both client and admin bookings. Config service setting need only affect client journeys and not admin

Solution
- bb_time_ranges_controller now checks if there is a configuration GeneralOptions.max_advance_datetime_days to override the service one


src/core/javascripts/services/general_options.js
```diff
@@ -32,7 +32,8 @@ angular.module('BB.Services').provider('GeneralOptions', function () {
         scroll_offset: 0,
         map_marker_icon: null,
         map_layout: 'default',
-        companyHasExternalBookings: false
+        companyHasExternalBookings: false,
+        maxAdvanceDatetimeDays: 100
```
MarkShakespeare: We should set the default to null, that way it will use the max_advance_datetime set in the service config by default

terencemutizwa: set the default to null

PR #805, opened 2017-06-17
Parent admin replacement of CORE-1447 

Glenn:
So - this pull request is interned to replace CORE-1447 

It's not 100% compelte - we can discuess next week, but there were a couple of core things I wanted to change:


Glenn:
One this I will add to this - is that there are still currently the data: {roles: []} on the states.  I want to get rid of these and have the states read their permissions from the nav menu config. I like that fact that is doesn't let you even route to pages that are not permissioned, and I understand you could have pages that are not in the nav menu specifically - but for which you also want to add some permissions checks - however I think the "permissions set" needs to be a seperate block like a const that can be later replced/read from an API/read form config etc.. and not in the javascript of the state definitions. So if it's not defined in the nav menu - then I think we need a new permissions set like that nav config :-)

